PR #3, opened 2019-06-13
2019-06-13 meeting minutes

piogor:
Hi, I am sending minutes from the meeting today.

### DIFF
--- a/meetings/2019-06-13/Agenda.md
+++ b/meetings/2019-06-13/Agenda.md
@@ -1,0 +1,26 @@
+Date: 2019-06-13
+
+1) Details explanation of the COSS and C4
+    -	Agree to use the COSS and C4 for the specification 
+
+2) Move gitlab:hardion/tango-rfc to github:tango-controls/tango-rfc
+    - Agree for the whole, partial or recreate from scratch
+    - Agree for the domain: “kernel-rfc.tango-controls.org” or other propositions
+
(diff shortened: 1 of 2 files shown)


hardion:
Would not it be better to use the wiki for the minutes?

piogor:
This may be reasonable. However, I do not have a strong opinion on it, as both approaches may serve their purpose :-) .
Having a meeting minutes along with sources may help in referencing. On the other side, it will create a PRs not directrly related to `valuable specification update`. I suggest to leave it as it is now and put it as a point to decide on the next meeting.

piogor:
Anyway, I would then need access to the wiki.

hardion:
You should have access to the wiki now.

hardion:
My preference would go for the wiki to not pollute the specification source.

piogor:
Ok, I have added the minutes as a wiki. One drawback is that GitHub wiki does not support hierarchical (tree, subpages) structure.

hardion:
Yes sure. In wiki style we can make pointers from the first page.
I will close the MR if you are agree.

piogor:
Yes, please close :)

Pobierz aplikację Outlook dla systemu Android<https://aka.ms/ghei36>

________________________________

hardion:
I have made a landing page for the wiki with a link to your minutes.
Thanks for your contribution!

Issue #5, opened 2019-06-14
The Command specification


hardion:
# Goals
Specify the concept of Tango Command

# Use Cases
There are several use cases:
* A command "Stop" can trigger the ending of a movement for a motor or an integration for a detector
* A command "Send" can take any SCPI command as argument and wait for the reply

# Specification
The specification has to be written and can be validated when all possible combination of command/argument/output can be specified.

piogor:
It seems to be done. I am closing the issue :-).

PR #7, opened 2019-06-27
The command RFC

hardion:
Related to issue #5 
This is just the beginning of the Command description.

TODO:
- [x] Change Editor
- [x] Use Cases
- [x] Remove comment that should be discussed in the slack channel
- [x] Add ALL argument types
- [x] Specific State, Status, Init Commands

### DIFF
--- a/3/README.md
+++ b/3/README.md
@@ -3,7 +3,8 @@ domain: rfc.tango-controls.org
 shortname: 3/Command
 name: Command
 status: draft
-editor: Vincent Hardion (vincent.hardion@maxiv.lu.se)
+editor: [Sergi Blanchi-Torné](mailto:sblanch@cells.es)
+contributors: [Vincent Hardion](mailto:vincent.hardion@maxiv.lu.se)
 ---
 
 This document describes the specification of the Tango Command model, a guideline for the implementer of the Tango Control System.
--- a/3/README.md
+++ b/3/README.md
@@ -99,12 +99,146 @@ Additionally, Command may be used for:
  dev-void-repr = DEVVOID | DevVoid
  dev-void-value = NULL
  
- 
  dev-double = dev-double-type dev-double-repr dev-double-value
  dev-double-type = 0x02
  dev-double-repr = DEVDOUBLE | DevDouble
  dev-double-value = ...
- ```
+
+ dev-boolean = dev-boolean-type dev-boolean-repr dev-boolean-value
+ dev-boolean-type = 0x03
+ dev-boolean-repr = DEVBOOLEAN | DevBoolean
+ dev-boolean-value = ...
+
+ dev-float = dev-float-type dev-float-repr dev-float-value
+ dev-float-type = 0x02
+ dev-float-repr = DEVFLOAT | DevFloat
+ dev-float-value = ...
+
+ dev-short = dev-short-type dev-short-repr dev-short-value
+ dev-short-type = 0x02
+ dev-short-repr = DEVSHORT | DevShort
+ dev-short-value = ...
+
+ dev-long = dev-long-type dev-long-repr dev-long-value
+ dev-long-type = 0x02
+ dev-long-repr = DEVLONG | DevLong
+ dev-long-value = ...
+
+ dev-long64 = dev-long64-type dev-long64-repr dev-long64-value
+ dev-long64-type = 0x02
+ dev-long64-repr = DEVLONG64 | DevLong64
+ dev-long64-value = ...
+
+ dev-string = dev-string-type dev-string-repr dev-string-value
+ dev-string-type = 0x02
+ dev-string-repr = DEVSTRING | DevString
+ dev-string-value = ...
+
+ dev-uchar = dev-uchar-type dev-uchar-repr dev-uchar-value
+ dev-uchar-type = 0x02
+ dev-uchar-repr = DEVUCHAR | DevUChar
+ dev-uchar-value = ... (octet)
+
+ dev-ushort = dev-ushort-type dev-ushort-repr dev-ushort-value
+ dev-ushort-type = 0x02
+ dev-ushort-repr = DEVUSHORT | DevUShort
+ dev-ushort-value = ...
+
+ dev-ulong = dev-ulong-type dev-ulong-repr dev-ulong-value
+ dev-ulong-type = 0x02
+ dev-ulong-repr = DEVULONG | DevULong
+ dev-ulong-value = ...
+
+ dev-ulong64 = dev-ulong64-type dev-ulong64-repr dev-ulong64-value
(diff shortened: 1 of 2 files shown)


srgblnch:
Once read the proposal, I have some comments:

a) In the second dot of the **goals** '_Expect a reply in any case_' I thing there would be 3 answers to a command (this part may share something with attributes):
- A command must have an input argument (that can be void) and return an output argument (that can also be void). But it may produce an exception or ends due to a timeout.

b) From the acronym of tango, the 'o' means 'object', so perhaps the meaning of the command could be quite similar to the meaning of 'method' in object oriented design, couldn't be? 

c) The last of the items in the goals, could lead to thing that we desire to use them like getters and setter. Perhaps this is not desired.

d) I the **user cases** one thing that can be said there is case of the sync and async commands

hardion:
a) I would make more clear in the specification part.
b) good point 
c) good point. I put it here because many devices used Command in this case for legacy reason. I will ad a disclaimer.
d) yes I can add one case

srgblnch:
b) You already say at the beginning of this 'goals' section "A Command in a Tango Controls represents an action". So action or procedure could be almost equivalent.
c) If I remember correctly, taco didn't have attribute so there was necessary to use setter and getter.

hardion:
b) I like your reference to the  Object oriented design
c) Yes, but I will try to not mention taco

srgblnch:
Thanks. May I add newer comments?

e) there is a TODO about string representation of the arguments in jive. Afaik, jive is interpreting the string in the gui to convert to the type of the argument. Using cpp or python one cannot send a string "3" instead of the number 3.

f) About the argument types. Was your intention to specify more that void and double, wasn't it?

hardion:
All comments are welcome of course ;-)
e)So API does not use type as string. Should we still enforce the String representation of type?
f) Yes, all the type should be described. This is my next push.

srgblnch:
e) I'm not sure I understand. An argument can be any of the tango types, including the DevString. But if it is a DevLong, one shouldn't pass the string representation of a number to be interpreted, should it?

g) at the end, in the 'Global Behaviour' section second and third item are duplicated, and fourth has a typo in 'ionly'.

h) I've searched in the current documentation in order to gain information about what has been wrote during the years. In [device server model](https://tango-controls.readthedocs.io/en/latest/development/device-api/device-server-model.html?highlight=command), there is a commands section. There it is mention that the commands can be synchronous or async. In out writing, in the goals, is it clear enough this two different execution types?

i) I think we miss the relation with the state of the devices to allow or deny a command execution, do we?

j) In the current implementation there is the 'expert' flag (I think it is display level). This I think is only used in some presentation layers as a filter, but could be nice to be the first step of an ACL. So perhaps it can be included as what we have description.

hardion:
Just added all type of Command Argument from the tango.idl.
The type index and the value specification are still to be done. 
For the value specification it involves to get more reading of the ABNF syntax.
k) Do we need the numerical representation of the type?

hardion:
Answer

> e) I'm not sure I understand. An argument can be any of the tango types, including the DevString. But if it is a DevLong, one shouldn't pass the string representation of a number to be interpreted, should it?

I just wonder if this important to have a standard name like DevEncoded. This is not really use in the Tango communication but more for the client like Jive. Anyway with the last push all type have their representation.

> g) at the end, in the 'Global Behaviour' section second and third item are duplicated, and fourth has a typo in 'ionly'.

Thanks I will check that.
 
> h) I've searched in the current documentation in order to gain information about what has been wrote during the years. In [device server model](https://tango-controls.readthedocs.io/en/latest/development/device-api/device-server-model.html?highlight=command), there is a commands section. There it is mention that the commands can be synchronous or async. In out writing, in the goals, is it clear enough this two different execution types?

This is very interesting. I will check if this is not only a client side implementation. 

> i) I think we miss the relation with the state of the devices to allow or deny a command execution, do we?

This will be part of the State Machine RFC.
 
> j) In the current implementation there is the 'expert' flag (I think it is display level). This I think is only used in some presentation layers as a filter, but could be nice to be the first step of an ACL. So perhaps it can be included as what we have description.

I will add that to the Command metadata.

srgblnch:
> k) Do we need the numerical representation of the type?

I see the 'dev-*-type' starts in 0x00, then 0x01, 0x02, 0x03, and the rest are left as 0x02. A enumeration could be nice. We, very often use in python if the type is one of a list and it is better to compare with values than strings.

This numbering could already exist so they we may continue with the same.

srgblnch:
l) I've seen in the branch, the 'README.md' in the root doesn't point to '3/README.md'

I've reasize on that because I've seen in the rfc-5 (about the properties) that is pointing to '5/README' without the 'md' extension.

hardion:
@srgblnch  Finally I simplified the argument specification to check only the literal form of the type and value. It's less precise but more accurate in order to compare different implementation.

3/README.md
```diff
@@ -0,0 +1,261 @@
...
+ dev-boolean-type = "DEVBOOLEAN" | "DevBoolean"
+ dev-boolean-value = BIT
+
+ dev-float = dev-float-type WSP dev-float-value
+ dev-float-type = "DEVFLOAT" | "DevFloat"
+ dev-float-value = floatnumber ; single-precision normalized numbers in IEC 60559
+
+ dev-short = dev-short-type WSP dev-short-value
+ dev-short-type = "DEVSHORT" | "DevShort"
+ dev-short-value = ["-"] DIGIT* ; In the range of "(2^15) - 1" to "-(2^15)".
```
lorenzopivetta, line 121: Should the range be lower-limit to upper-limit, e.g. "In the range of -(2^15) to (2^15)-1"

3/README.md
```diff
@@ -0,0 +1,261 @@
...
+ dev-ushort-type = "DEVUSHORT" | "DevUShort"
+ dev-ushort-value = DIGIT* ; In the range of "0" to "(2^16) - 1".
+
+ dev-ulong = dev-ulong-type WSP dev-ulong-value
+ dev-ulong-type = "DEVULONG" | "DevULong"
+ dev-ulong-value = DIGIT* ; In the range of "0" to "(2^32) - 1".
+
+ dev-ulong64 = dev-ulong64-type WSP dev-ulong64-value
+ dev-ulong64-type = "DEVULONG64" | "DevULong64"
+ dev-ulong64-value = DIGIT* ; In the range of "0" to "(2^64) - 1".
```
piogor, line 149: Awhat about this way: `dev-ulong64-value = unsignednteger  ; In range of`...

david-erb:
Will the details about data types be moved to a separate RFC?

hardion:
Thanks all.

Have a nice week end

PR #29, opened 2019-07-10
Fix links & add Sergi Blanchi-Torné as RFC-3 editor

bourtemb:


bourtemb:
Sorry, I was not sure I should fork for such minor changes...
The created branch should be deleted after the merge to avoid to pollute the rfc master repo with too many branches...

bourtemb:
Well, we can discuss whether we link directly to the README.md file or to the directory...
I think linking to the directory is fine because there might be several files involved in some situations.
It looks like @gwen-soleil committed by mistake to tango-controls/rfc draft branch.
If you want I can revert that commit and the previous 2 other commits which introduced the link issue in the README file (adding Gwen as Property Editor).
Then I could cancel #29 (this PR) and remove the branch which was created for it and we will update the editor list from other Pull Requests coming from forks?
Or we fix the links in this PR... As you wish!

piogor (approved):
Thanks, Reynald. I will merge this PR then I will prepare a new one for new editors.

Issue #6, opened 2019-06-14
The Property Specification


hardion:
# Goals
The specification of Property aims to define a standard structure and type to use in any implementation of Tango.

Additionally, it aims to:

* Provide the initial information when one starts a Tango Device

* Be usable as keeping global variable in the Tango control system

* Provide meta data complementing any data in Tango

# Use Cases
There are many use cases for the usage of a Property:

* To connect a real network equipment a Property can define the target IP address

* To change the representation of an information in order to be more user friendly a Property can define a new data format for the graphical user interface to convert the raw data.

hardion:
Note that a preliminary draft of this specification has been pushed. The editor/contributors can take it as a base or overwrite it.
My original idea was to define simply the property semantically as a key / value and derive this term for all configuration like attribute property, configuration.

PR #30, opened 2019-07-10
WIP for the Properties RFC-5

bourtemb:
Related to issue #6

### DIFF
--- a/5/README.md
+++ b/5/README.md
@@ -7,6 +7,7 @@ editor: Gwenaëlle Abeillé `<at synchrotron-soleil.fr - gwenaelle.abeille>`
 contributors:
   - Vincent Hardion `<at maxiv.lu.se - vincent.hardion>`
   - Reynald Bourtembourg `<at esrf.fr - bourtemb>`
+  - Andy Gotz <andy.gotz@esrf.fr>
 ---
 
 This document describes the Property, a Tango concept representing an element of configuration in order to customise a Tango element. This document describes version 1.0 of the Property.
--- a/5/README.md
+++ b/5/README.md
@@ -16,7 +16,7 @@ See also: 2/Device, 6/Database, 4/Attribute
 
 ## Preamble
 
-Copyright (c) 2019 MAX IV Laboratory.
+Copyright (c) 2019 Tango Controls
 
 This Specification is free software; you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation; either version 3 of the License, or (at your option) any later version. This Specification is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details. You should have received a copy of the GNU General Public License along with this program; if not, see <http://www.gnu.org/licenses>.
 
@@ -28,10 +28,10 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 
  A Property is a persisted item of Tango that is mainly used for configuration purposes. 4 types of properties exist:
 
- *  Class property: a property that is attached to a Device Class (cf RFC-9). This property is accessible from all devices of the class
- *  Device property: a property that is attached to a Device (RFC-2). This property is accessible from only one device
- *  Attribute property: a property that is attached to an Attribute (RFC-4).
- *  Free property: a property to persist data that is not related to any Tango class, device or attribute.
+ *  Class Property: a Property related to a Device Class (cf RFC-2).
+ *  Device Property: a Property related to a Device (RFC-2).
+ *  Attribute Property: a Property related to an Attribute (RFC-4).
+ *  Free Property: a Property defined for the entire Control System (RFC-1) i.e not related to any Class, Device or Attribute.
 
 ### Goals
 
@@ -41,7 +41,7 @@ Additionally, it aims to:
 
 * Provide the initial information when one starts a Tango Device
 
-* Be usable as keeping global variable in the Tango control system
+* Be usable as keeping global variable in the Tango Control System
 
 * Provide meta data complementing any data in Tango
 
@@ -54,27 +54,26 @@ There are many use cases for the usage of a Property:
 
 * To change the representation of an information in order to be more user friendly a Property can define a new data format for the graphical user interface to convert the raw data.
 
-* To configure some alarms on attribute's value.
+* To configure some alarms on Attribute value.
 
-* To memorize the attribute value each time it changes.
+* To memorize the Attribute value each time it changes.
 
 ## Specification
 
 A Tango Property is a strict definition of a pair of key/value
 * The Property SHALL have one key, called Property Name
 * The Property SHALL have one value, which could be empty, called Property Value
-* If the device is started upon a Tango Database, Class, Device, Attribute and Free Properties MAY be persisted into the Tango Database (RFC6)
-* If the device is started without a Tango Database, the device and class properties MAY be persisted in a file (but no support for the attribute properties). 
-* All device and class properties MAY be directly modified and accessed through the database device or its file
-* Attribute properties SHOULD be modified and accessed at any time through the device object (RFC-2)
-* All default attribute and device properties are OPTIONAL
-* The attribute, class and device properties MAY be defined in the device code and they MAY be OPTIONAL OR REQUIRED (also called mandatory properties).
-* TODO: system properties
-* If the device is started upon a Tango Database, the class and device properties changes are historized in the Tango Database (RFC-6)
-* When a error occurs (i.e. network issue, Tango Database timeout...), a DevFailed exception MUST be thrown (RFC-x for DevFailed?)
-* All device and class properties MUST be loaded at init command or device start-up (RFC-8)
-* A device or class property MAY have an OPTIONAL default value. 
-	* The device property value loading flow in pseudo-code is:
+* If the Device is started upon a Tango Database, Class, Device, Attribute and Free Properties MAY be persisted into the Tango Database (RFC6)
+* If the Device is started without a Tango Database, the device and class properties MAY be persisted in a file (but no support for the attribute properties). 
+* All Device and Class Properties MAY be directly modified and accessed through the database device or its file (TODO: to be moved to Request/Reply RFC)
+* Attribute Properties SHOULD be modified and accessed at any time through the Device object (RFC-2) (TODO: to be moved to Request/Reply RFC)
+* A Property Value MAY have a default value
+* A Mandatory Property is a Property for which the Property Value SHALL NOT be null and SHALL not have a default value
+* If the Device is started upon a Tango Database, the Class and Device Properties changes are historized in the Tango Database (RFC-6) (TODO: To be moved to Tango Database RFC)
+* When a error occurs (i.e. network issue, Tango Database timeout...), a DevFailed exception MUST be thrown (RFC-x for DevFailed/Common types?) (TODO: To be moved to Request-Reply RFC)
+* All Device and Class Properties MAY be loaded at init command or device start-up (RFC-8)
+* A Device or Class Property MAY have a default value. 
+* The Device Property Value loading flow in pseudo-code is (TODO to be moved to RFC-8):
 	```
 	property-value = getDevicePropertyFromTangoDBorFile(device-name, property-name)
 	if(property-value is empty)
@@ -82,27 +81,43 @@ A Tango Property is a strict definition of a pair of key/value
 	if(property-value is empty && default-value is not empty)
   		property-value = default-value
   	```
-	* The class property valueloading flow is:
+	* The Class property value loading flow is:
 	```
 	property-value = getClassPropertyFromTangoDBorFile(class-name, property-name)
 	if(property-value is empty && default-value is not empty)
   		property-value = default-value
   	```
-* All device and class properties MAY be modified and accessed through the database device when using a Tango Database (RFC-6)
-* A Tango device MUST manage the following default device properties, see RFC-cache and RFC-logging:
+* The precedence of Property Value for a given Property Name SHALL follow this order of preference:
+  1) the last Device Property Value
+  2) the last Class Property Value
+  3) the default Device Property Value
+  4) the default Class Property Value
+  
+  Meaning that:
+  * The last Property Value has precedence over the default Property Value
+  * The Device Property Value has precedence over the Class Property
+* All Device and Class Properties MAY be modified and accessed through the database device when using a Tango Database (RFC-6)
+* A Tango Device MUST manage the following default Device Properties, which are called System Properties, see RFC-cache:
 
 | Name | Default value |  Description |
 | --- | --- | --- |
-| poll_ring_depth | "" | ? |
+| poll_ring_depth | 10 |  	Polling buffer depth |
+| cmd_poll_ring_depth | "" | List of command polling history depth |
+| attr_poll_ring_depth | "" | List of attribute polling history depth |
+| poll_old_factor | 4 | Data too old factor |
 | min_poll_period | "" | Minimum polling period for attributes and commands |
 | cmd_min_poll_period | "" | List of command minimum polling periods |
-| cmd_poll_ring_depth | "" | List of command polling history depth |
 | attr_min_poll_period | "" | List of attribute minimum polling periods |
-| attr_poll_ring_depth | "" | List of attribute polling history depth |
-| polled_attr | "" | List of polled attribute's names|
-| logging_target | "" | Device logging target, a file or a device. ex: "file:/tmp/file.log", "device:log/device/logger.1" |
-| logging_level | "" | Device logging level (FATAL, ERROR, WARN, INFO, DEBUG, OFF) |
-| logging_rft | "" | ? |
+| polled_attr | "" | List of polled attribute's names |
+
+* A Tango Device MUST manage the following default Device Properties, which are called Logging Properties, see RFC-logging:
+
+| Name | Default value |  Description |
+| --- | --- | --- |
+| logging_level | WARN | initial Device logging level (FATAL, ERROR, WARN, INFO, DEBUG, OFF) |
+| logging_path | "/tmp/tango-<logging name>" or "C:/tango-<logging name> (Windows)" | Logging file path |
+| logging_rft | 20480 (20 Mega bytes) | Logging rolling file threshold in bytes |
+| logging_target | "" | Initial Device logging target, a file or a device. ex: "file:/tmp/file.log", "device:log/device/logger.1" |
 
 * An attribute MUST manage the following default attribute properties (TODO: move this to specifics RFCs: Attribute Configuration properties / Attribute Event properties / Attribute Properties?):
 	* Attribute Configuration properties :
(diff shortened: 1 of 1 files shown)


bourtemb:
I attach a presentation I made for Icalepcs 2015 Tango workshop on Tango properties [TangoProperties.zip](https://github.com/tango-controls/rfc/files/3378625/TangoProperties.zip)
This could provide some ideas on this very wide topic.

5/README.md
```diff
@@ -45,23 +53,72 @@ There are many use cases for the usage of a Property:
...
+		* String **__root_att** default value "Not specified"
+		* String **enum_labels**
+		* String **__value**
+
 
 ### Naming convention
 * The Property's Name SHALL use the following convention:
 ``` ABNF
 property-name = 1*VCHAR
 ```
```
t-b, line 169: From my testing with Jive I've found that I can add a property with an empty name as well. And I can also just use a space (` `) as name. There are even no checks for duplicated names so two properties with empty names are possible as well. So I think this is more like `property-name = 0*CHAR`.

Looking into the database schema, https://github.com/tango-controls/TangoDatabase/blob/master/create_db_tables.sql.in, the name is limited to 255 characters so we can write it as.

```ABNF
property-name = 0*255CHAR
```

5/README.md
```diff
@@ -45,23 +53,72 @@ There are many use cases for the usage of a Property:
...
+* The property name is case insensitive
+* The Property value MUST use the following convention:
+        single value that may contain any character
+        or an array with carriage return
+        special values: "NaN", "inf"
```
t-b: So the property values are something like

```ABNF
propery-value = CHAR / 1*CHAR CR / "NaN" / "inf"
```

?

5/README.md
```diff
@@ -45,23 +53,72 @@ There are many use cases for the usage of a Property:
...
+* All default attribute and device properties are OPTIONAL
+* The attribute, class and device properties MAY be defined in the device code and they MAY be OPTIONAL OR REQUIRED (mandatory)
+* TODO memorized attributes, system properties
+* TODO : properties history
+* TODO: error management
+* All device and class properties MUST be loaded at init command or device start-up
+* If a device property does not exist, the device MUST load the class property having the same name
+* All device and class properties MAY be modified and accessed through the database device when using a Tango Database (RFC-6)
+* A Tango device MUST manage the following default device properties: cmd_min_poll_period, min_poll_period, attr_min_poll_period, poll_ring_depth, cmd_poll_ring_depth, attr_poll_ring_depth, polled_attr, logging_target, logging_level, logging_rft (TODO: description of each one)
+* An attribute MUST manage the following default attribute properties (TODO: descriptions):
```
mliszcz: It it Property RFC or Attribute RFC which shall describe things like *max_alarm*?
Yesterday I've started to work on Attribute RFC. You may check if you want, but I still need do some rework and testing before submitting a PR:
https://github.com/mliszcz/rfc/blob/fix-28-attribute-spec/4/README.md

5/README.md
```diff
@@ -45,23 +53,72 @@ There are many use cases for the usage of a Property:
...
+* If the device is started without a Tango Database, the device and class properties MAY be persisted in a file (but no support for the attribute properties) 
+* All device and class properties MAY be directly modified and accessed through the database device or its file
+* Attribute properties SHOULD be modified and accessed at any time through the device object (RFC-2)
+* All default attribute and device properties are OPTIONAL
+* The attribute, class and device properties MAY be defined in the device code and they MAY be OPTIONAL OR REQUIRED (mandatory)
+* TODO memorized attributes, system properties
+* TODO : properties history
+* TODO: error management
+* All device and class properties MUST be loaded at init command or device start-up
+* If a device property does not exist, the device MUST load the class property having the same name
```
lorenzopivetta: The current flow is:
1 - check if class property exist, load it
2 - check if default value exist, load it
3 - check if value is defined in database, load it
This guarantees that, whenever defined, default value will override class value, and db value will override default value (and class as well).
Propose to reword line 74 something like:
"For each Property, the device checks and extracts Property value from Class, default device value and database defined value, with increasing priority"

bourtemb: > The current flow is:
> 1 - check if class property exist, load it
> 2 - check if default value exist, load it
> 3 - check if value is defined in database, load it

A more correct way to express it would be:
The current flow is:
1 - check if class property exist, load it, **else** check if default value exist, load it
2 - (executed in any case) check if value is defined in database, load it

The way you wrote it suggested that the default property would override the class property (at least this is the way I interpreted it)

bourtemb: > The way you wrote it suggested that the default property would override the class property (at least this is the way I interpreted it)

And it looks like this is the way you interpreted it too.. This is not the case in the current implementations.  A default value does not override a class property.

bourtemb: So, starting from your sentence proposal, I would change it to:
_For each Property, the device checks and extracts default device value, Property value from Class and database defined value, with increasing priority._

lorenzopivetta: You're right. Had a quick look and missed the "else". But what about "_For each Property, the device checks and extracts Property value from default device value, Class and database defined value, with increasing priority_" ?

bourtemb: > But what about "For each Property, the device checks and extracts Property value from default device value, Class and database defined value, with increasing priority" ?

Looks better to me indeed.

5/README.md
```diff
@@ -1,14 +1,17 @@
...
 status: raw
-editor: Vincent Hardion (vincent.hardion@maxiv.lu.se)
+editor: Gwenaëlle Abeillé `<at synchrotron-soleil.fr - gwenaelle.abeille>`
+contributors:
+  - Vincent Hardion `<at maxiv.lu.se - vincent.hardion>`
+  - Reynald Bourtembourg `<at esrf.fr - bourtemb>`
 ---
 
-This document describes the Property,  a Tango concept representing an element of configuration in order to customise a Tango element. This document describes version 1.0 of the Property.
+This document describes the Property, a Tango concept representing an element of configuration in order to customise a Tango element. This document describes version 1.0 of the Property.
```
lorenzopivetta: Tango "element" is sort of undefined concept...

5/README.md
bourtemb: > From my testing with Jive I've found that I can add a property with an empty name as well. And I can also just use a space (` `) as name. There are even no checks for duplicated names so two properties with empty names are possible as well. So I think this is more like `property-name = 0*CHAR`.

This would not be true if we consider properties defined from a file (without Tango Database).
POGO does some checks on the property name and won't accept an empty property name and would remove any space. POGO also forbids device property names with some special characters like '%' for instance which could be interpreted as a wildcard by a Database engine.

> Looking into the database schema, https://github.com/tango-controls/TangoDatabase/blob/master/create_db_tables.sql.in, the name is limited to 255 characters so we can write it as.

Indeed, there is this limitation by default.
This is also defined in cppTango for instance:
https://github.com/tango-controls/cppTango/blob/tango-9-lts/cppapi/server/tango_const.h.in#L221

> 
> ```abnf
> property-name = 0*255CHAR
> ```

As we said this morning during the teleconf, this might be true with the current jive implementation (not true when using POGO or a file DB). We will make a note that the current implementation allows this kind of property name (empty name of with spaces and tabs) but this is due to issues with the current implementation. empty property names, property names with tabs, spaces and some special characters (we need to find the list) should be forbidden.

bourtemb: Here is the method which does the check in POGO:
https://github.com/tango-controls/pogo/blob/Pogo-Release-9.6.23/org.tango.pogo.pogo_gui/src/org/tango/pogo/pogo_gui/tools/Utils.java#L404

As I understand this code, only alphanumeric characters are allowed [a-z][A-Z][0-9] and the underscore.
The property shall start with a letter.
An attribute property (and only an attribute property) can start with an underscore.

@srgblnch , @hardion , this could be useful for #7: this [POGO piece of code](https://github.com/tango-controls/pogo/blob/Pogo-Release-9.6.23/org.tango.pogo.pogo_gui/src/org/tango/pogo/pogo_gui/tools/Utils.java#L404) (combined with a slightly deeper analysis) also highlights that "State" and "Status" command names are reserved. Command names seem to be case sensitive because command names "state" and "status" seem to be accepted by POGO.
I've done a quick test generating a C++ device server with a command named "state" returning a DevState and it seems that this use case is not well supported. Pogo generated the source code, but created a method 
`Tango::DevState MyClassName::state()`

AtkPanel did not recognize this "state" command. tg_devtest (jive test device) listed "state" and "State" as 2 commands available on the device but executing the "state" command from tg_devtest did not execute Tango::DevState MyClassName::state() method, but MyClass::dev_state() was executed instead, as for the "State" command.

Similar behaviour when adding a command named "status".

So it would probably be good to specify that "state" and "status" should be reserved|forbidden too as command names.

andygotz: This is indeed a deficiency of Pogo and should be avoided. To be added to RFC on Commands, Attributes and State.

5/README.md
andygotz: I would say:

property-value = 1*[1*VCHAR | "nan" | "inf"] 

where VCHAR is the visible character set and nan and inf are case in-sensitive. The CR is part of the implementation in the database. I am not sure it belongs in the property value definition.

5/README.md
andygotz: I propose to create multiple RFC's for all the different types of attribute properties:

1. Attribute Configuration properties
2. Attribute Event properties
3. Attribute Properties

@andygotz + @bourtemb

5/README.md
andygotz: Element could be any field of a Tango control system like Tango Device, Command, Attribute, System. or free value i.e. not a Tango object. What about this formulation:
```suggestion
This document describes the Property, a Tango concept representing one or more values to configure one of the following Tango elements namely Tango Device, Class, Attribute, System or a not associated to any Tango element, so called free properties. This document describes version 1.0 of the Property.
```

5/README.md
andygotz: What about replacing "database defined value" with "device specific value". Or we should say that  the Device specific value can be defined in multiple places e.g. database, file or in the code (aie!).

andygotz:
@bourtemb just pointed out there is a default property in the Class and Device supported by pogo. 

TODO make sure this is captured in the RFC

5/README.md
lorenzopivetta: Sounds good to me

gwen-soleil: ok also for me.

5/README.md
gwen-soleil: `property-value = 1*[1*VCHAR | "nan" | "inf"]` does not respect the ABNF standard for me (https://tools.ietf.org/html/rfc5234 or https://en.wikipedia.org/wiki/Augmented_Backus%E2%80%93Naur_form). If the CR is part of the implementation in the database, how can we define arrays?

5/README.md
lorenzopivetta: The way I was reading this is:
default device Property value = if specified in the Pogo dialog... this ends up in the XXXClass.cpp file
default Class Property value = if specified in the Pogo dialog... this ends up in the XXXClass.cpp file
database Class Property value = if Class Property defined in the Tango db... 
database device Property value = if device Property defined in the Tango db...
Not sure about your reference to "file", does this refers to running a device with db on file? Isn't this just a special case of database?

5/README.md
gwen-soleil: Since the specifications were not there at the time of implementation, I think it is normal that we fall into those kind of inconsistencies. So I think we should compromise on a specification even though the current implementation does not reflect 100% the specification. I propose:
`property-name = 1*255ALPHA`

5/README.md
gwen-soleil: To simplify the reading, I would propose to add pseudo-code. Something like:

- A device or class property MAY have an OPTIONAL default value
- The device property loading flow in pseudo-code is:

```
property-value = getDevicePropertyFromTangoDBorFile(device-name, property-name)
if(property-value is empty)
  property-value = getClassPropertyFromTangoDBorFile(class-name, property-name)
if(property-value is empty && default-value is not empty)
  property-value = default-value
```

- The class property loading flow property-name is:

```
property-value = getClassPropertyFromTangoDBorFile(class-name, property-name)
if(property-value is empty && default-value is not empty)
  property-value = default-value`
```

5/README.md
hardion: > @srgblnch , @hardion , this could be useful for #7: this [POGO piece of code](https://github.com/tango-controls/pogo/blob/Pogo-Release-9.6.23/org.tango.pogo.pogo_gui/src/org/tango/pogo/pogo_gui/tools/Utils.java#L404) (combined with a slightly deeper analysis) also highlights that "State" and "Status" command names are reserved. Command names seem to be case sensitive because command names "state" and "status" seem to be accepted by POGO.
> I've done a quick test generating a C++ device server with a command named "state" returning a DevState and it seems that this use case is not well supported. Pogo generated the source code, but created a method
> `Tango::DevState MyClassName::state()`

Indeed a good point. The draft of the device RFC already define to have a State and Status attributes. Would it make sense to define what is a State and a Status command in the Command RFC and make mandatory the State and Status command in the Device RFC?

5/README.md
```diff
@@ -22,7 +26,12 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 
 ## Tango Property Specification
 
- A Property is designed to represent any information of configuration in Tango.
+ A Property is a persisted item of Tango that is mainly used for configuration purposes. 4 types of properties exist:
+
+ *  Class property: a property that is attached to a Device Class (cf RFC-9). This property is accessible from all devices of the class
+ *  Device property: a property that is attached to a Device (RFC-2). This property is accessible from only one device
```
bourtemb: > This property is accessible from all devices of the class
> This property is accessible from only one device
I find these sentences a bit confusing... Technically, nothing prevents another device or a client to have access to a device property value defined in the DB or even to the default value of a device property.

5/README.md
bourtemb: > Here is the method which does the check in POGO:
> https://github.com/tango-controls/pogo/blob/Pogo-Release-9.6.23/org.tango.pogo.pogo_gui/src/org/tango/pogo/pogo_gui/tools/Utils.java#L404
> 
> As I understand this code, only alphanumeric characters are allowed [a-z][A-Z][0-9] and the underscore.
> The property name shall start with a letter.
> An attribute property (and only an attribute property) can start with an underscore.

So if we follow the rules defined by code in POGO, I think we would get something like that for the property names in abnf:

```
alphanum = ALPHA / DIGIT 
underscore = %x5F
device_property_name = 1*ALPHA *(alphanum / underscore)
class_property_name = 1*ALPHA *(alphanum / underscore)
free_property_name = 1*ALPHA *(alphanum / underscore)
attribute_property_name = 1*(ALPHA / underscore) *(alphanum / underscore)
```

bourtemb: And if we take into account that the maximum property name length is 255 we would get something like:
```
alphanum = ALPHA / DIGIT 
underscore = %x5F
device_property_name = 1*1ALPHA 0*254(alphanum / underscore)
class_property_name = 1*1ALPHA 0*254(alphanum / underscore)
free_property_name = 1*1ALPHA 0*254(alphanum / underscore)
attribute_property_name = 1*1(ALPHA / underscore) 0*254(alphanum / underscore)
```

5/README.md
```diff
@@ -45,23 +54,138 @@ There are many use cases for the usage of a Property:
...
 
-## Specification
-
-A Tango Property is a strict definition of a pair of key/value
-* configuration are represented in the form of properties.
-* data are represented in the form of attributes.
-* actions are represented in the form of commands.
-
-Its model can be represented as a defined tree which each elements are from a defined types: Class, Device, Property, Attribute, Command following the rules below:
+* To memorize the Attribute value each time it changes.
```
lorenzopivetta, line 59: Does this means "marking an attribute as memorized" ? I'd rewrite this as "To mark an attribute as memorized, and store the value each time it changes".

bourtemb: @lorenzopivetta , I agree it is currently a bit ambiguous. I would not use "To mark an attribute as memorized" because this part is done in the code, not in the DB.
I would reformulate to something like:
* to memorize the Attribute value of a Memorized Attribute when it changes

The idea here was to talk about the role of the __value attribute property.

5/README.md
```diff
@@ -45,23 +54,138 @@ There are many use cases for the usage of a Property:
...
+	```
+	property-value = getClassPropertyFromTangoDBorFile(class-name, property-name)
+	if(property-value is empty && default-value is not empty)
+  		property-value = default-value
+  	```
+* The precedence of Property Value for a given Property Name SHALL follow this order of preference:
+  1) the last Device Property Value
+  2) the last Class Property Value
+  3) the default Device Property Value
+  4) the default Class Property Value
```
bourtemb, line 94: This is not completely true...
Here are the results of some tests I did:

| device prop. value in DB | class prop. value in DB | device prop. default value | device prop. mandatory in DB | class prop. default value | Prop value at the end of init_device |
| --- | --- | --- | --- | --- | --- |
| not defined | not defined | not defined | TRUE | not defined | Error |
| not defined | not defined | not defined | TRUE | Class Property Default Value | Class Property Default Value |
| Device Property Value in DB | not defined | not defined | TRUE | Class Property Default Value | Device Property Value in DB |
| Device Property Value in DB | not defined | not defined | TRUE | not defined | Device Property Value in DB |
| not defined | not defined | not defined | FALSE | not defined | empty string. 0.0 for double (not initialized in the code so, depends on compiler?) |
| not defined | not defined | not defined | FALSE | Class Property Default Value | Class Property Default Value |
| Device Property Value in DB | not defined | not defined | FALSE | Class Property Default Value | Device Property Value in DB |
| Device Property Value in DB | not defined | not defined | FALSE | not defined | Device Property Value in DB |
| not defined | not defined | Device Property Default Value | FALSE | not defined | Device Property Default Value |
| not defined | not defined | Device Property Default Value | FALSE | Class Property Default Value | Class Property Default Value |
| Device Property Value in DB | not defined | Device Property Default Value | FALSE | Class Property Default Value | Device Property Value in DB |
| Device Property Value in DB | not defined | Device Property Default Value | FALSE | not defined | Device Property Value in DB |
| not defined | Class Property Value in DB | not defined | TRUE | not defined | Class Property Value in DB |
| not defined | Class Property Value in DB | not defined | TRUE | Class Property Default Value | Class Property Value in DB |
| Device Property Value in DB | Class Property Value in DB | not defined | TRUE | Class Property Default Value | Device Property Value in DB |
| Device Property Value in DB | Class Property Value in DB | not defined | TRUE | not defined | Device Property Value in DB |
| not defined | Class Property Value in DB | not defined | FALSE | not defined | Class Property Value in DB |
| not defined | Class Property Value in DB | not defined | FALSE | Class Property Default Value | Class Property Value in DB |
| Device Property Value in DB | Class Property Value in DB | not defined | FALSE | Class Property Default Value | Device Property Value in DB |
| Device Property Value in DB | Class Property Value in DB | not defined | FALSE | not defined | Device Property Value in DB |
| not defined | Class Property Value in DB | Device Property Default Value | FALSE | not defined | Class Property Value in DB |
| not defined | Class Property Value in DB | Device Property Default Value | FALSE | Class Property Default Value | Class Property Value in DB |
| Device Property Value in DB | Class Property Value in DB | Device Property Default Value | FALSE | Class Property Default Value | Device Property Value in DB |
| Device Property Value in DB | Class Property Value in DB | Device Property Default Value | FALSE | not defined | Device Property Value in DB |

This is valid only for C++.
I found some of the results quite surprising. 
For instance, if you have a class property with a default value and a mandatory device property with the same name with no value at all in the DB, the device server will use the class default value in this case.
So, the property is defined as mandatory in DB but there is nothing in the DB.

When you have a class property with a default value and a device property with a default value, and nothing defined in the DB for these property, then the device server will use the class property default value.
With POGO, you cannot generate code

bourtemb: Pogo gives an error when you try to generate code for a Java device server with a class property and a device property having the same property name.
For PythonHL, Pogo can generate the code in this use case, but the behaviour is different than in C++ since it is assigning twice a (different) value to the same class member. The first time for the class property and the second time for the device property. So at the end, it is always a device property.

```
    # ----------------
    # Class Properties
    # ----------------

    prop_with_default_class_and_prop_val = class_property(
        dtype='DevString', default_value="Default Class Property Value"
    )

    # -----------------
    # Device Properties
    # -----------------

    prop_with_default_class_and_prop_val = device_property(
        dtype='DevString', default_value="Default Device Property Value"
    )
```

bourtemb:
A new PR has been created for this RFC/5 (#59). 
Please Re-open #30 if you think we should rather continue to work on the first created PR.

Issue #28, opened 2019-07-10
The Attribute specification


mliszcz:
# Goals

Specify the concept of Tango Attribute.

# Use cases

Some examples of possible use cases (source [[tango-doc]]):

* A device associated with a motor has a position attribute expressed in mm.
* A device associated with a thermocouple has a temperature attribute expressed in Celsius (or any another suitable unit).

[tango-doc]: https://tango-controls.readthedocs.io/en/latest/development/device-api/ds-guideline/device-server-guidelines.html?highlight=attribute#purpose-of-an-attribute

PR #32, opened 2019-07-15
Add specification for the Attribute model (#28)

mliszcz:
This PR adds specification for the Attribute model (see #28).

TODO / WIP:
* forwarded attributes

Open points for discussion:
* which RFC shall provide spec for polling? Maybe we should have a dedicated RFC. Polling is configured at device level and can be also applied to commands,
* which RFC shall describe attribute properties (e.g. "min_alarm")? I've seen some work for such properties already done in Property RFC,
* should we have a dedicated RFC for aliases? To define attribute alias I need concepts like "full object name" and "attribute proxy".

### DIFF
--- a/4/README.md
+++ b/4/README.md
@@ -68,6 +68,9 @@ An Attribute MUST have associated following static metadata:
   (one of *SCALAR*, *SPECTRUM*, *IMAGE*),
 * *writable*, an enumeration describing the write access to the Attribute
   (one of *READ*, *WRITE*, *READ_WRITE*, *READ_WITH_WRITE*).
+  * An Attribute can be read from, if *writable* is one of *READ*, *READ_WRITE*,
+    *READ_WITH_WRITE*.
+  * An Attribute can be written to, if *writable* is one of *WRITE*, *READ_WRITE*.
 * *display level*, an enumeration describing visibility level of the Attribute
   (one of *EXPERT*, *OPERATOR*).
 
@@ -78,10 +81,6 @@ An Attribute MUST have associated following static metadata:
 > applications. Also note that the Attribute name SHOULD contain at least one
 > letter and SHOULD NOT start with a digit.
 
-An Attribute can be read from, if *writable* is one of *READ*, *READ_WRITE*,
-*READ_WITH_WRITE*.
-
-An Attribute can be written to, if *writable* is one of *WRITE*, *READ_WRITE*.
 
 [^1]: In current implementation it is possible to define more than one
 Attribute with the same name, but only one can be accessed externally.
--- a/4/README.md
+++ b/4/README.md
@@ -1,0 +1,248 @@
+---
+domain: rfc.tango-controls.org
+shortname: 4/Attribute
+name:Attribute
+status: draft
+editor: [Sergi Blanchi-Torné](mailto:)
+contributors: [Michal Liszcz](mailto:michal.liszcz@s2innovation.com)
+---
+
+This document describes the Tango Attribute model specification version 5.0.
+
+
+## Preamble
+
+Copyright (c) 2019 Tango Controls
+
+This Specification is free software; you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation; either version 3 of the License, or (at your option) any later version. This Specification is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details. You should have received a copy of the GNU General Public License along with this program; if not, see <http://www.gnu.org/licenses>.
+
+This Specification is a [free and open standard](http://www.digistan.org/open-standard:definition) and is governed by the Digital Standards Organization's [Consensus-Oriented Specification System](http://www.digistan.org/spec:1/COSS).
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119).
+
+
+## Tango Attribute specification
+
+This specification is intended to formally document the Tango Attribute static
+model. Runtime representation of Attribute in conforming Tango implementations
+MUST follow this specification.
+
+
+### Goals
+
+An Attribute is a Tango concept representing a physical quantity of a device.
+The main purpose of an Attribute is to provide read and (optionally) write
+access to this quantity.
+
+
+### Use Cases
+
+Some example use cases of an Attribute are:
+* an Attribute can represent a position of a motor device,
+* an Attribute can represent a temperature of a thermocouple device.
+
+
+## Specification
+
+An Attribute has:
+* A set of static metadata that constitute to Attribute's definition,
+* A set of dynamically configurable properties, e.g. for alarming purposes,
+* A set of runtime parameters describing Attribute's value at given point
+  in time.
+
+### Attribute definition
+
+The static metadata is part of the Attribute definition. It MUST be defined
+before an Attribute is created by any Tango implementation and MUST NOT change
+at runtime after the Attribute is initialized.
+
+An Attribute MUST have associated following static metadata:
+* *name*, a string identifying the Attribute.
+  It MUST be unique [^1] among all Attributes of a particular device.
+  It MAY contain any character (including multi-byte characters)
+  except that it MUST NOT contain `\0` character.
+  It MAY be empty,
+* *data type*, an enumeration describing the type of the data (X/DataTypes),
+* *data format*, an enumeration describing the dimension of the data
+  (one of *SCALAR*, *SPECTRUM*, *IMAGE*),
+* *writable*, an enumeration describing the write access to the Attribute
+  (one of *READ*, *WRITE*, *READ_WRITE*, *READ_WITH_WRITE*).
+* *display level*, an enumeration describing visibility level of the Attribute
+  (one of *EXPERT*, *OPERATOR*).
+
+An Attribute can be read from, if *writable* is one of *READ*, *READ_WRITE*,
+*READ_WITH_WRITE*.
+
+An Attribute can be written to, if *writable* is one of *WRITE*, *READ_WRITE*.
+
+[^1]: In current implementation it is possible to define more than one
+Attribute with the same name, but only one can be accessed externally.
+
+If *data format* is *SPECTRUM*,
+an Attribute MUST have associated following additional static metadata:
+* *max dim x*, an integer describing the maximum allowed number of data
+  elements.
+  The Attribute MUST be able to store up to *max dim x* data elements.
+  It MUST be greater than 0.
+
+If *data format* is *IMAGE*,
+an Attribute MUST have associated following additional static metadata:
+* *max dim x*, an integer describing the maximum allowed number of data
+  elements in dimension X.
+  The Attribute MUST be able to store up to *max dim x* data elements
+  in the X dimension.
+  It MUST be greater than 0,
+* *max dim y*, an integer describing the maximum allowed number of data
+  elements in dimension Y.
+  The Attribute MUST be able to store up to *max dim y* data elements
+  in the Y dimension.
+  It MUST be greater than 0.
+
+If *writable* is *READ_WITH_WRITE*,
+an Attribute MUST have associated following additional static metadata:
+* *writable attribute name*, a string describing the *name* of associated
+  writable attribute.
+  It MUST point to an attribute.
+  The Attribute pointed to MUST have *writable* set to one of *WRITE*,
+  *READ_WRITE*.
+
+If *data type* is *ENUM*,
+an Attribute MAY have associated following additional static metadata:
+* *enum labels*, a list of strings describing textual representation
+  of enumerated values.
+  Its elements MAY contain any character (including multi-byte characters)
+  except that they MUST NOT contain `\0` character.
+  It MAY be empty.
+
+TODO
+These are associated with the device in DB but impact the attribute.
+TBD where we want to document such properties.
+* *polled*,
+* *polling period*.
+
+
+#### Memorized attribute
+
+An Attribute MUST have associated following static metadata:
+* *memorized*, a flag describing whether *set value* is stored in the database
+  and restored during Attribute initialization,
+* *write hardware at init*, a flag describing whether *set value* is
+  written to the Attribute during initialization or *init* call.
+  It is effective only if *memorized* is also set.
+
+To set *memorized*, following constraints MUST be satisfied:
+* *data format* MUST be *SCALAR*,
+* *data type* MUST NOT be *STATE* or *ENCODED*,
+* *writable* MUST NOT be *READ* or *READ_WITH_WRITE*.
+
+If *memorized* is set and if database is being used,
+the *set value* MUST be persisted in the *__value* property upon each write
+to the Attribute.
+
+If *memorized* is set and *__value* property is other than "Not used yet"
+(default):
+* only during Attribute initialization (at startup):
+  * *set value* MUST be set with a value stored in *__value* property,
+* during Attribute initialization (at startup) or during an *init* command
+  execution, if *write hardware at init* is also set:
+  * value stored in *set value* MUST be written to the Attribute.
+
+#### Forwarded attribute
+
+TODO
+
+
+### Attribute properties
+
+An Attribute MAY have associated following dynamic metadata in form of
+Properties (see RFC 5/Property):
+* *description*, providing textual information about the Attribute,
+* *label*, providing textual label to use as Attribute's name,
+* *unit*, providing textual representation of Attribute's unit,
+* *standard unit*, describing the conversion factor to transform Attribute's
+  value into S.I units. It MUST be a valid numerical value,
+* *display unit*, describing the conversion factor to transform Attribute's
+  value into value usable in GUIs. It MUST be a valid numerical value,
+* *format*, describing how to convert Attribute's data to a textual
+  representation. It MUST comply to
+  [`printf`](http://man7.org/linux/man-pages/man3/printf.3.html)
+  format string specification,
+* *min value*, describing the minimum value of Attribute's *read value*.
+  It MUST be a vaild numerical value.
+  It MUST be defined only for Attributes with numerical *data type*.
+  It MUST be defined only for writable Attributes.
+  It MUST be lower than *max value*,
+* *max value*, describing the maximum value of Attribute's *read value*.
+  It MUST be a vaild numerical value.
+  It MUST be defined only for Attributes with numerical *data type*.
+  It MUST be defined only for writable Attributes.
+  It MUST be greater than *max value*,
+* *min alarm*, describing the threshold value of Attribute's *read value*
+  below which the attribute is considered as having alarm *quality*.
+  It MUST be a vaild numerical value.
+  It MUST be defined only for Attributes with numerical *data type*.
+  It MUST be lower than *max alarm*,
+* max alarm*, describing the threshold value of Attribute's *read value*
+  above which the attribute is considered as having alarm *quality*.
+  It MUST be a vaild numerical value.
+  It MUST be defined only for Attributes with numerical *data type*.
+  It MUST be greater than *min alarm*,
+* *min warning*, is deprecated,
+* *max warning*, is deprecated,
+* *delta val*, describing the maximum difference between Attribute's
+  *read value* and *set value* after *delta t* time period,
+  above which the attribute is considered as having alarm *quality*.
+  It MUST be defined only for writable Attributes,
+* *delta t*, describing the time period after which the difference between
+  *read value* and *set value* must be lower than *delta val*.
+  Otherwise the attribute is considered as having alarm *quality*.
+  It MUST be defined only for writable Attributes,
+  It MUST be defined if *delta val* is also defined.
+* TODO: describe event-related properties,
+* *rel change*,
+* *abs change*,
+* *period*, 
+* *archive rel change*,
+* *archive abs change*,
+* *archive period*,
+
+
+### Attribute runtime parameters
+
+At given point in time an Attribute MUST have associated:
+* *quality*, an enumeration describing the state *read value*
+  (one of *VALID*, *INVALID*, *ALARM*, *CHANGING*, *WARNING*),
+* *read value*, an object representing the value of the Attribute.
+  It MUST conform to *data format* and *data type*,
+* *read dim x*, an integer describing the number of data elements in
+  *read value* in X dimension.
+  If *data format* is *SCALAR*, it MUST be either 0 or 1.
+  If *data format* is *SPECTRUM* or *IMAGE*, it MUST be between 0 and
+  *max dim x*.
+* *read dim y*, an integer describing the number of data elements in
+  *read value* in Y dimension.
+  If *data format* is *SCALAR* or *SPECTRUM*, it MUST be 0.
+  If *data format* is *IMAGE*, it MUST be between 0 and *max dim y*.
+
+Additionally, if *writable* is *WRITE* or *READ_WRITE*,
+at given point in time an Attribute MUST have associated:
+* *set value*, an object representing the value set to the Attribute.
+  It MUST conform to *data format* and *data type*,
+* *write dim x*, an integer describing the number of data elements in
+  *set value* in X dimension,
+* *write dim y*, an integer describing the number of data elements in
+  *set value* in Y dimension.
+
+
+### Attribute aliases
+
+TODO
+An Attribute MAY have associated *aliases*.
+
+
+### TODO
+
+Formal specification of Attribute static model is given below:
+```ABNF
+attribute = "whatever"
+```
--- a/4/README.md
+++ b/4/README.md
@@ -343,7 +343,7 @@ and *USER* events without the need to configure the polling for the Attribute.
 
 Formal specification of Attribute name is given below:
 ```ABNF
-attribute-name = *attribute-name-char
+attribute-name = 1*attribute-name-char
 attribute-name-char = %d48-57 / %d65-90 / %d97-122 / "_" ; 0-9 / A-Z / a-z / _
 
 full-attribute-name = device-name "/" attribute-name
--- a/4/README.md
+++ b/4/README.md
@@ -150,7 +150,22 @@ If *memorized* is set and *__value* property is other than "Not used yet"
 
 #### Forwarded attribute
 
-TODO
+An Attribute can be a forwarded attribute. A forwarded Attribute MUST have
+associated *__root_att* property.
+It MUST be defined before attribute initialization at startup.
+It MUST NOT be changed in runtime.
+It MUST point to an existing attribute in another device.
+
+Allowed syntax for *__root_att* is specified below as `<full-attribute-name>`.
+
+Each read and write request to a forwarded Attribute MUST be passed
+to the Attribute pointed by *__root_att*.
+
+Each change to properties or to other metadata associated with a forwarded
+Attribute MUST be reflected in the Attribute pointed by *__root_att*.
+
+Each change to properties or to other metadata associated with the Attribute
+pointed by *__root_att* MUST be reflected in the forwarded Attribute.
 
 
 ### Attribute properties
--- a/4/README.md
+++ b/4/README.md
@@ -3,7 +3,7 @@ domain: rfc.tango-controls.org
 shortname: 4/Attribute
 name:Attribute
 status: draft
-editor: [Sergi Blanchi-Torné](mailto:)
+editor: [Sergi Blanchi-Torné](mailto:sblanch@cells.es)
 contributors: [Michal Liszcz](mailto:michal.liszcz@s2innovation.com)
 ---
 
--- a/4/README.md
+++ b/4/README.md
@@ -61,7 +61,8 @@ An Attribute MUST have associated following static metadata:
   It MUST be unique [^1] among all Attributes of a particular device.
   It MAY contain any character (including multi-byte characters)
   except that it MUST NOT contain `\0` character.
-  It MAY be empty,
+  It MAY be empty.
+  See `<attribute-name>` specification below,
 * *data type*, an enumeration describing the type of the data (X/DataTypes),
 * *data format*, an enumeration describing the dimension of the data
   (one of *SCALAR*, *SPECTRUM*, *IMAGE*),
@@ -236,13 +237,31 @@ at given point in time an Attribute MUST have associated:
 
 ### Attribute aliases
 
-TODO
-An Attribute MAY have associated *aliases*.
+A *full attribute name* consists of *device name* and *attribute name*
+separated by "/" character.
 
+An Attribute MAY have associated *alias*, a string which can be used
+in place of *full attribute name* to address the Attribute.
+An Attribute MUST NOT have more than one *alias*.
 
-### TODO
+An *alias* MUST be stored in the database.
 
-Formal specification of Attribute static model is given below:
+An *alias* MUST be unique.
+
+An *alias* MAY contain any character (including multi-byte characters) except
+that it MUST NOT contain any of `/`, ` ` (space), `#`, `:`, `->`.
+It MAY be empty.
+See `<attribute-alias>` specification below.
+
+### Attribute naming schema
+
+Formal specification of Attribute name is given below:
 ```ABNF
-attribute = "whatever"
+attribute-name = *attribute-name-char
+attribute-name-char = %x01-FF ; OCTET except %x00
+
+full-attribute-name = device-name "/" attribute-name
+
+attribute-alias = *attribute-alias-char
+attribute-alias-char = OCTET ; except %x00 "/" " " "#" ":" "->"
 ```
--- a/4/README.md
+++ b/4/README.md
@@ -171,7 +171,7 @@ pointed by *__root_att* MUST be reflected in the forwarded Attribute.
 ### Attribute properties
 
 An Attribute MAY have associated dynamic metadata in form of
-Properties (see RFC 5/Property).
+Properties (see [RFC 5/Property](/5)).
 
 General properties:
 * *description*, providing textual information about the Attribute,
@@ -282,6 +282,8 @@ at given point in time an Attribute MUST have associated:
 
 A *full attribute name* consists of *device name* and *attribute name*
 separated by "/" character.
+
+See [RFC 2/Device](/2) for the definition of `<device-name>`.
 
 An Attribute MAY have associated *alias*, a string which can be used
 in place of *full attribute name* to address the Attribute.
--- a/4/README.md
+++ b/4/README.md
@@ -62,7 +62,6 @@ at runtime after the Attribute is initialized.
 An Attribute MUST have associated following static metadata:
 * *name*, a string identifying the Attribute.
   It MUST be unique [^1] among all Attributes of a particular device.
-  It MAY contain any character (including multi-byte characters).
   See `<attribute-name>` specification below,
 * *data type*, an enumeration describing the type of the data (X/DataTypes),
 * *data format*, an enumeration describing the dimension of the data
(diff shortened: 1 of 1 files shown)


hardion:
> Open points for discussion:
> 
> * which RFC shall provide spec for polling? Maybe we should have a dedicated RFC. Polling is configured at device level and can be also applied to commands,

Polling mechanism can be described in the RFC/Cache System

> * which RFC shall describe attribute properties (e.g. "min_alarm")? I've seen some work for such properties already done in Property RFC,

IHMO Attribute RFC is the right place since alarm system concern 99% the  Tango attribute (some side effect on the Device State which can be describe in the Device RFC)

> * should we have a dedicated RFC for aliases? To define attribute alias I need concepts like "full object name" and "attribute proxy".

Just make a reference to the Device RFC.

mliszcz:
Thanks @hardion!
> Polling mechanism can be described in the RFC/Cache System

I have added a note regarding polling and referenced RFC/Cache system.

> IHMO Attribute RFC is the right place since alarm system concern 99% the Tango attribute (some side effect on the Device State which can be describe in the Device RFC)

Already documented.

> Just make a reference to the Device RFC.

Done.

srgblnch:
Cześć @mliszcz !

Sorry for the delay in my task.I'm back on line. I'll use letter to tag comments I do about the text so then we can reference them in an easier way.

a) In the goals section, there are some more attribute types* than 'read only' and 'read write', I've never used but there is the possibility to have 'write only' and 'read with write'. You later on mention this but don't fit with the first lines of the rfc.
\* Not the same type than datatype.

b) like in the commands rfc, I suggest to reference the Object Oriented paradigm to describe what an attribute mean in Tango. Just the meaning of the name of Tango does a direct reference to the Object paradigm. I think this reference help then in the User Cases section because it already fits in a common view.

c) In the attribute definition, the name may have more restrictions (like don't contain slashes I think).

d) Continuing with the attribute definition for the name, I think it would be nice if, apart to describe what we already have in the implementation, guide to encourage or discourage unpleasant behaviors like for me it is the empty name.

I'll do another read soon in order to think more in a refining of the rfc.

Thanks for the effort!

./Sergi

mliszcz:
Thanks @srgblnch!

> a) In the goals section, there are some more attribute types* than 'read only' and 'read write',

My intention was to not go into details in the goals section. Shall I reword the goals and include all attribute access types?

> b) like in the commands rfc, I suggest to reference the Object Oriented paradigm

Sure, I'll add this information.

> c) In the attribute definition, the name may have more restrictions (like don't contain slashes I think).

I've checked with c++ server and python client. Both `\` and `/` are fine. Maybe some combination of letters is not allowed - I need to have a look at the source code.

```cpp
rw_scalar_doubleAttrib():Attr("~`\1!@#$%^&*()_+{}[]:;'|\\<,>.?/ǢǣǮȌ-=", Tango::DEV_DOUBLE, Tango::READ_WRITE) {};
```

```python
In [6]: filter(lambda a: '!@#' in a.name, dp.attribute_list_query())[0].name
Out[6]: "~`\x01!@#$%^&*()_+{}[]:;'|\\<,>.?/\xc7\xa2\xc7\xa3\xc7\xae\xc8\x8c-="

In [8]: dp.read_attribute("~`\x01!@#$%^&*()_+{}[]:;'|\\<,>.?/\xc7\xa2\xc7\xa3\xc7\xae\xc8\x8c-=").value
Out[8]: 4.6464171418034e-310
```

> d) Continuing with the attribute definition for the name, I think it would be nice if, apart to describe what we already have in the implementation, guide to encourage or discourage unpleasant behaviors like for me it is the empty name.

I've just checked the command and property RFCs. Seems that they also had some problems with naming and ended up with a simple:

```abnf
command-name = 1*VCHAR
```

Maybe we shall do the same for the Attribute? I'd like to avoid giving guidelines in a specification, because the guidelines are subjective.

bourtemb:
> > c) In the attribute definition, the name may have more restrictions (like don't contain slashes I think).
> 
> I've checked with c++ server and python client. Both `\` and `/` are fine. Maybe some combination of letters is not allowed - I need to have a look at the source code.
> 
> ```c++
> rw_scalar_doubleAttrib():Attr("~`\1!@#$%^&*()_+{}[]:;'|\\<,>.?/ǢǣǮȌ-=", Tango::DEV_DOUBLE, Tango::READ_WRITE) {};
> ```
> 
> ```python
> In [6]: filter(lambda a: '!@#' in a.name, dp.attribute_list_query())[0].name
> Out[6]: "~`\x01!@#$%^&*()_+{}[]:;'|\\<,>.?/\xc7\xa2\xc7\xa3\xc7\xae\xc8\x8c-="
> 
> In [8]: dp.read_attribute("~`\x01!@#$%^&*()_+{}[]:;'|\\<,>.?/\xc7\xa2\xc7\xa3\xc7\xae\xc8\x8c-=").value
> Out[8]: 4.6464171418034e-310
> ```

It is true that the library lets you create some attributes with these strange characters but POGO doesn't. Please refer to :
https://github.com/tango-controls/pogo/blob/0887143c96f95576a242f69d7f953db72ba93eea/org.tango.pogo.pogo_gui/src/org/tango/pogo/pogo_gui/tools/Utils.java#L404-L436

In the original Tango philosophy, Device servers programmers had to use POGO to define their attributes. Then the dynamic attribute feature got implemented and it seems like no checks were added to the library. I think it is a mistake/inconsistency.

> 
> > d) Continuing with the attribute definition for the name, I think it would be nice if, apart to describe what we already have in the implementation, guide to encourage or discourage unpleasant behaviors like for me it is the empty name.
> 
> I've just checked the command and property RFCs. Seems that they also had some problems with naming and ended up with a simple:
> 
> ```abnf
> command-name = 1*VCHAR
> ```

Please note that what is proposed in the property RFC is not definitive. It is a draft. We simply didn't have enough time to come up with a more precise abnf definition yet.

> 
> Maybe we shall do the same for the Attribute? I'd like to avoid giving guidelines in a specification, because the guidelines are subjective.

mliszcz:
Hi @srgblnch 
I think I resolved:
b) - just one extra sentence referring to the OOP in the *Goals* section
and c) + d) - added a note with RECOMMENDations for attribute naming - @bourtemb can you also comment if it's ok for you now? I think use of Pogo should not be mandatory (e.g. in Python HL there is almost no boilerplate code to generate).

4/README.md
```diff
@@ -0,0 +1,346 @@
...
+The static metadata is part of the Attribute definition. It MUST be defined
+before an Attribute is created by any Tango implementation and MUST NOT change
+at runtime after the Attribute is initialized.
+
+An Attribute MUST have associated following static metadata:
+* *name*, a string identifying the Attribute.
+  It MUST be unique [^1] among all Attributes of a particular device.
+  It MAY contain any character (including multi-byte characters)
+  except that it MUST NOT contain `\0` character.
+  It MUST not not be "Status" or "State", which are reserved names.
```
bourtemb: State and Status are actually valid attribute names so I think this sentence is not correct... It is just that the device server programmer can not create additional attributes having these names since this would contradict the first point above: "It MUST be unique among all Attributes of a particular device".

4/README.md
```diff
@@ -0,0 +1,346 @@
...
+
+Some example use cases of an Attribute are:
+* an Attribute can represent a position of a motor device,
+* an Attribute can represent a temperature of a thermocouple device.
+
+
+## Specification
+
+An Attribute has:
+* A set of static metadata that constitute to Attribute's definition,
```
bourtemb: This sentence does not sound very English to me.
A set of static metadata that constitute Attribute's definition ?
A set of static metadata that contribute to Attribute's definition ?
...?

4/README.md
```diff
@@ -0,0 +1,346 @@
...
+before an Attribute is created by any Tango implementation and MUST NOT change
+at runtime after the Attribute is initialized.
+
+An Attribute MUST have associated following static metadata:
+* *name*, a string identifying the Attribute.
+  It MUST be unique [^1] among all Attributes of a particular device.
+  It MAY contain any character (including multi-byte characters)
+  except that it MUST NOT contain `\0` character.
+  It MUST not not be "Status" or "State", which are reserved names.
+  It MAY be empty.
```
bourtemb: It MAY be empty but SHOULD NOT be empty.
I am not convinced we should even put a sentence like "It MAY be empty" when we know that it won't work correctly in this use case. I would remove this sentence to avoid confusion.

4/README.md
```diff
@@ -0,0 +1,346 @@
...
+
+The static metadata is part of the Attribute definition. It MUST be defined
+before an Attribute is created by any Tango implementation and MUST NOT change
+at runtime after the Attribute is initialized.
+
+An Attribute MUST have associated following static metadata:
+* *name*, a string identifying the Attribute.
+  It MUST be unique [^1] among all Attributes of a particular device.
+  It MAY contain any character (including multi-byte characters)
+  except that it MUST NOT contain `\0` character.
```
bourtemb: I find this '\0' detail quite confusing for the C Binding client implementation for instance...

4/README.md
```diff
@@ -0,0 +1,346 @@
...
+  except that they MUST NOT contain `\0` character.
+  It MAY be empty.
+
+
+#### Memorized attribute
+
+An Attribute MUST have associated following static metadata:
+* *memorized*, a flag describing whether *set value* is stored in the database
+  and restored during Attribute initialization,
+* *write hardware at init*, a flag describing whether *set value* is
```
bourtemb: I would use "memorized set value" instead of "set value" here.

4/README.md
```diff
@@ -0,0 +1,346 @@
...
+  written to the Attribute during initialization or *init* call.
+  It is effective only if *memorized* is also set.
+
+To set *memorized*, following constraints MUST be satisfied:
+* *data format* MUST be *SCALAR*,
+* *data type* MUST NOT be *STATE* or *ENCODED*,
```
bourtemb: Indeed Attr:set_memorized() method seems to forbid memorized attributes of type STATE:
https://github.com/tango-controls/cppTango/blob/f3efe02db1ef1bfd5347901f0e7b5e20aba478da/cppapi/server/attrdesc.cpp#L665-L675

I think this is a mistake in the current implementation. Attributes of type DevState (other than attribute named "State") were not there at the beginning when attributes were created in Tango. It looks like the test in Attr::set_memorized() has not been updated when the support of DevState user attributes was added.
In theory, there is no reason not to support STATE memorized attributes and we can even see that some other methods are ready for that, like Device_3Impl::write_attributes_in_db for instance:
https://github.com/tango-controls/cppTango/blob/tango-9-lts/cppapi/server/device_3.cpp#L2296-L2382

So, thank you @mliszcz for this discovery. We should fix that in cppTango (and add a test to test State memorized attributes).

4/README.md
```diff
@@ -0,0 +1,346 @@
...
+* *writable* MUST NOT be *READ* or *READ_WITH_WRITE*.
+
+If *memorized* is set and if database is being used,
+the *set value* MUST be persisted in the *__value* property upon each write
```
bourtemb: *__value* property  => *__value* attribute property

4/README.md
```diff
@@ -0,0 +1,346 @@
...
+to the Attribute.
+
+If *memorized* is set and *__value* property is other than "Not used yet"
```
bourtemb: *__value* property  => *__value* attribute property

4/README.md
```diff
@@ -0,0 +1,346 @@
...
+(default):
+* only during Attribute initialization (at startup):
+  * *set value* MUST be set with a value stored in *__value* property,
```
bourtemb: *__value* property  => *__value* attribute property

4/README.md
```diff
@@ -0,0 +1,346 @@
...
+* during Attribute initialization (at startup) or during an *init* command
+  execution, if *write hardware at init* is also set:
+  * value stored in *set value* MUST be written to the Attribute.
+
+#### Forwarded attribute
+
+An Attribute can be a forwarded attribute. A forwarded Attribute MUST have
+associated *__root_att* property.
```
bourtemb: *__root_att* property => *__root_att* attribute property

4/README.md
```diff
@@ -0,0 +1,346 @@
...
+* *standard unit*, describing the conversion factor to transform Attribute's
+  value into S.I units. It MUST be a valid numerical value,
+* *display unit*, describing the conversion factor to transform Attribute's
+  value into value usable in GUIs. It MUST be a valid numerical value,
+* *format*, describing how to convert Attribute's data to a textual
+  representation. It MUST comply to
+  [`printf`](http://man7.org/linux/man-pages/man3/printf.3.html)
+  format string specification,
+* *min value*, describing the minimum value of Attribute's *read value*.
+  It MUST be a vaild numerical value.
```
bourtemb: vaild  => valid
(there are several other occurences of this typo below)

4/README.md
```diff
@@ -0,0 +1,346 @@
...
+  It MUST be defined only for Attributes with numerical *data type*.
```
bourtemb, line 188: MUST => MAY ?
It can be unspecified

4/README.md
```diff
@@ -0,0 +1,346 @@
...
+  It MUST be defined only for writable Attributes.
+  It MUST be lower than *max value*,
```
bourtemb: if *max value* is specified

4/README.md
```diff
@@ -0,0 +1,346 @@
...
+* *unit*, providing textual representation of Attribute's unit,
+* *standard unit*, describing the conversion factor to transform Attribute's
+  value into S.I units. It MUST be a valid numerical value,
+* *display unit*, describing the conversion factor to transform Attribute's
+  value into value usable in GUIs. It MUST be a valid numerical value,
+* *format*, describing how to convert Attribute's data to a textual
+  representation. It MUST comply to
+  [`printf`](http://man7.org/linux/man-pages/man3/printf.3.html)
+  format string specification,
+* *min value*, describing the minimum value of Attribute's *read value*.
```
bourtemb: I would rather put:
*read value* => *set value*

4/README.md
```diff
@@ -0,0 +1,346 @@
...
+  It MUST be a vaild numerical value.
+  It MUST be defined only for Attributes with numerical *data type*.
+  It MUST be defined only for writable Attributes.
+  It MUST be lower than *max value*,
+* *max value*, describing the maximum value of Attribute's *read value*.
```
bourtemb: Same here, I think *max value* is especially important for the *set/write value*

4/README.md
```diff
@@ -0,0 +1,346 @@
...
+  It MUST be a vaild numerical value.
+  It MUST be defined only for Attributes with numerical *data type*.
+  It MUST be defined only for writable Attributes.
+  It MUST be greater than *max value*.
```
bourtemb: It MUST be greater than *max value*. => It MUST be greater than *min value*.

4/README.md
```diff
@@ -0,0 +1,346 @@
...
+  It MUST be a vaild numerical value.
+  It MUST be defined only for Attributes with numerical *data type*.
+  It MUST be lower than *max alarm*,
+* max alarm*, describing the threshold value of Attribute's *read value*
+  above which the attribute is considered as having alarm *quality*.
+  It MUST be a vaild numerical value.
+  It MUST be defined only for Attributes with numerical *data type*.
+  It MUST be greater than *min alarm*,
+* *min warning*, is deprecated,
+* *max warning*, is deprecated,
```
bourtemb: Where did you find the information that *min warning* and *max warning* are deprecated?
The documentation does not mention that: https://tango-controls.readthedocs.io/en/latest/development/device-api/attribute-alarms.html?highlight=min_warning#the-level-alarms

4/README.md
```diff
@@ -0,0 +1,346 @@
...
+* *delta val*, describing the maximum difference between Attribute's
+  *read value* and *set value* after *delta t* time period,
+  above which the attribute is considered as having alarm *quality*.
+  It MUST be defined only for writable Attributes,
+* *delta t*, describing the time period after which the difference between
```
bourtemb: Can we specify the unit of this time period?

4/README.md
```diff
@@ -0,0 +1,346 @@
...
+Properties for event reporting purposes:
+* *rel change*, *abs change*, *archive rel chagne*, *archive abs chagne*,
+  describing the threshold values for relative and absolute change
+  in Attribute's *read value* above which
+  a *CHANGE* event or *ARCHIVE* event MUST be reported.
+  It MUST be either a valid numerical value
+  or a pair of valid numerical values separated by a `,`.
+  If one value is specified, it MUST be used for both negative
+  and positive change.
+  If two values are specified, the first MUST be used for netative change
```
bourtemb: netative  => negative

4/README.md
```diff
@@ -0,0 +1,346 @@
...
+
+Properties for event reporting purposes:
+* *rel change*, *abs change*, *archive rel chagne*, *archive abs chagne*,
+  describing the threshold values for relative and absolute change
+  in Attribute's *read value* above which
+  a *CHANGE* event or *ARCHIVE* event MUST be reported.
+  It MUST be either a valid numerical value
+  or a pair of valid numerical values separated by a `,`.
+  If one value is specified, it MUST be used for both negative
+  and positive change.
```
bourtemb, line 245: It would be good to give more specifications about the relative change. How is it expressed?
What does a relative change of 1 means?

4/README.md
```diff
@@ -0,0 +1,346 @@
...
+  If two values are specified, the first MUST be used for netative change
+  and the second MUST be used for positive change.
+  It MUST be defined only for Attributes with numerical *data type*,
+* *period*, *archive period*, describing the minimum time period
+  in milliseconds after which a *PERIODIC* or *ARCHIVE* event MUST be reported,
+  regardless of Attribute's *read value*. If *period* is not specified,
+  a default value of 1000 ms is used,
+  It MUST be defined only for Attributes with numerical *data type*.
```
bourtemb: >  It MUST be defined only for Attributes with numerical *data type*.

I don't think this is true for the *period* and *archive period*

4/README.md
```diff
@@ -0,0 +1,346 @@
...
+
+See [Attribute events section](#attribute-events) for more details.
+
+```abnf
+rel-change = change
+abs-change = chagne
```
bourtemb: chagne => change (same a bit below)

4/README.md
```diff
@@ -0,0 +1,346 @@
...
+
+archive-rel-change = change
+archive-abs-change = chagne
+
+number = [ "-" ] 1*DIGIT [ "." ] *DIGIT
+change = number [ "," number ]
+
+period = [ "-" ] 1*DIGIT
```
bourtemb: negative period?

4/README.md
```diff
@@ -0,0 +1,346 @@
...
+### Attribute aliases
+
+A *full attribute name* consists of *device name* and *attribute name*
+separated by "/" character.
+
+See [RFC 2/Device](/2) for the definition of `<device-name>`.
+
+An Attribute MAY have associated *alias*, a string which can be used
+in place of *full attribute name* to address the Attribute.
+An Attribute MUST NOT have more than one *alias*.
```
bourtemb, line 313: I think this is a limitation of the current system with jive but I don't see why we could not have several aliases for the same attribute. How did you come to this conclusion? Did you find something in the doc or in the source code or was it simply by trying to define an attribute alias with jive?

4/README.md
```diff
@@ -0,0 +1,346 @@
...
+
+An *alias* MUST be stored in the database.
+
+An *alias* MUST be unique.
```
bourtemb: in the sense that a given alias name can only be associated to a unique attribute?

4/README.md
```diff
@@ -0,0 +1,346 @@
...
+* *CHANGE*,
+* *ARCHIVE*,
+* *PIPE*,
+* *DATA READY*,
+* *ATTR CONF*,
+* *USER*,
+
+In order to send *PERIODIC*, *CHANGE*, *ARCHIVE*, *PIPE* and *DATA READY*
+events, an attribute must be polled. See
+[Attribute polling section](#attribute-events) for more details.
```
bourtemb: This is not true. Events can be pushed by code. The polling is not mandatory in this case.

4/README.md
```diff
@@ -0,0 +1,346 @@
...
+An Attribute can be polled. A polling for the Attribute can be configured
+in Device with *polled attr* Property. See [RFC XX/Cache system](/XX) for
+more details.
+
+
+### Attribute naming schema
+
+Formal specification of Attribute name is given below:
+```ABNF
+attribute-name = *attribute-name-char
```
bourtemb: I think we should impose at least one character here because it does not work properly in the current system with empty attribute names.

bourtemb (commented):
Thanks @mliszcz for this very interesting work. This actually helps to find some issues with the current code.

4/README.md
mliszcz: Proposed fix: remove "It MUST not not be "Status" or "State", which are reserved names."

4/README.md
mliszcz: Proposed fix: "A set of static metadata that constitute Attribute's definition".

By "attribute definition" I meant all the parameters (name, type, etc...) you configure statically in Pogo (or pass in `Tango::Attr` constructor) and cannot be changed later at runtime.

4/README.md
mliszcz: Below is a note that there should be at least one character in attribute's name. Proposed fix: I'll remove "It MAY be empty."

4/README.md
mliszcz: You can't have \0 (since it's end of string), but you can have  other bytes like \1, \2, ... But - such names will not be accepted by Pogo (which expects only ASCII letters and numbers).

I can remove the note about \0 ("except that it MUST NOT contain `\0` character.")

Do we want to remove the "any character" part as well? - "It MAY contain any character (including multi-byte characters)". Even though it will not work with Pogo, I think non-English letters in attribute names are still a valid use case.

4/README.md
mliszcz: I'll do as you propose

4/README.md
mliszcz: So, do you propose to remove this constrains from the RFC? Or shall I add a note here that this will be changed in the future?

4/README.md
mliszcz: Will change as proposed.

4/README.md
mliszcz: Will change.

4/README.md
mliszcz: The header says that the Attribute MAY have properties listed below.
By "it MUST ..." I meant "if defined, it MUST ...".

Should I explicit add "if defined" to each requirement?

4/README.md
bourtemb: > You can't have \0 (since it's end of string), but you can have other bytes like \1, \2, ... But - such names will not be accepted by Pogo (which expects only ASCII letters and numbers).
> 
> I can remove the note about \0 ("except that it MUST NOT contain `\0` character.")

I don't know what would be the best. When reading it I just found it confusing:
> *name*, a **string** identifying the Attribute.
> it MUST NOT contain `\0` character.

A C string contains a `\0` at the end by definition so this seemed a bit contradictory to me.
 
> Do we want to remove the "any character" part as well? - "It MAY contain any character (including multi-byte characters)". Even though it will not work with Pogo, I think non-English letters in attribute names are still a valid use case.

Non-English letters/multi-byte characters in attribute names can be a valid use case indeed.  There might be some precautions needed for this to work correctly though. I think character set encoding configuration is important in this case on the database side , on the CORBA side and/or on the OS side.
I don't know whether some users are already doing it. This use case is not tested in cppTango I think.

4/README.md
mliszcz: Proposed fix: "... describing the time period (**in milliseconds**) after which the difference between ..."

4/README.md
bourtemb: > 
> 
> So, do you propose to remove this constrains from the RFC? Or shall I add a note here that this will be changed in the future?

This constraint on the type (STATE or ENCODED) might not exist in the latest version of JTango.

JTango does not (no longer) support(s) spectrum or image memorized attributes mainly due to performance issues:

https://github.com/tango-controls/JTango/blob/jtango-9-lts/server/src/main/java/org/tango/server/attribute/AttributeImpl.java#L482-L492

If an implementation manages to support these kinds of memorized attributes in an efficient manner, I would not forbid it.

We could add a note about what kind of memorized attributes are currently supported by JTango and cppTango.

In any case, what is important here is the concept of memorized attribute.

4/README.md
bourtemb: > Should I explicit add "if defined" to each requirement?

That would be one way to make it less confusing indeed, I think! 
Or maybe having something like:

If defined:
- It MUST ...
- It MUST ...

I am just sharing ideas (I am not the editor ;-) )

4/README.md
mliszcz: Proposed fix:
It MUST be expressed as a precentage change relative to the value reported in the previous event, e.g. a relative change of 1 will generate an event when:
```
(current value - previous value) / previous value > 1%
```

4/README.md
mliszcz: Correct! Proposed fix: remove "It MUST be defined only for Attributes with numerical *data type*."

4/README.md
mliszcz: Correct. It's either positive or 0 (default / not set). Proposed fix: `period = 1*DIGIT`

4/README.md
mliszcz: This is how it is implemented in the database:
https://github.com/tango-controls/TangoDatabase/blob/master/DataBase.cpp#L6150-L6155

Upon alias definition, you clear all existing aliases. Also see:

```
In [2]: db = tango.Database()

In [3]: db.put_attribute_alias('sys/tg_test/1/long_scalar', 'first_alias_for_test_long_scalar')

In [4]: db.get_attribute_alias_list('*')
Out[4]: DbDatum(name = '*', value_string = ['first_alias_for_test_long_scalar', 'rw_scalar_double-x', '', 'debian8_attr_alias'])

In [5]: db.put_attribute_alias('sys/tg_test/1/long_scalar', 'second_alias_for_test_long_scalar')

In [6]: db.get_attribute_alias_list('*')
Out[6]: DbDatum(name = '*', value_string = ['second_alias_for_test_long_scalar', 'rw_scalar_double-x', '', 'debian8_attr_alias'])
```

4/README.md
mliszcz: It must be unique across the whole database.
Please see DataBase::db_put_attribute_alias.
* First, we reject the new alias if this alias is already in use for another attribute
* Then we delete all aliases for the current attribute:
* And we finally insert the new alias

Effectively, the `alias` column is unique.

4/README.md
bourtemb: You're right indeed! What you wrote is correct! I just don't understand why we have this limitation for the moment. I thought it might be useful in some situations to have several aliases pointing to the same attribute.
Maybe this is not needed.

4/README.md
mliszcz: Proposed fix: replace above text with:

Events can be send automatically by a polling mechanism, automaticallyby the Tango system or manually from the device server code.

The polling mechanism can send *PERIODIC*, *CHANGE* and *ARCHIVE* events.
In order to send events via the polling mechanism, the polling for the Attribute MUST be configured in the Device with *polled attr* Property.
See "Attribute properties" section for conditions upon which the events are sent.
See "RFC XX/Cache system" for more details on polling configuration.

The Tango system MUST send *ATTR CONF* event if Attribute configuration is changed.
See "Attribute properties" section for a list of modifiable attribute Properties.

The device server code can manually send *CHANGE*, *ARCHIVE*, *DATA READY* and *USER* events without the need to configure polling for the Attribute.

bourtemb: > Proposed fix: replace above text with:
> 
> Events can be send automatically by a polling mechanism, automaticallyby the Tango system or manually from the device server code.

can be send => can be sent
automaticallyby  => automatically by 

> 
> The polling mechanism can send _PERIODIC_, _CHANGE_ and _ARCHIVE_ events.
> In order to send events via the polling mechanism, the polling for the Attribute MUST be configured in the Device with _polled attr_ Property.

I would just write something like:
"In order to send events via the polling mechanism, the polling for the Attribute MUST be enabled"
without going into more details. The polling RFC can give all the details on how to configure/start the polling for a given attribute.

> The device server code can manually send _CHANGE_, _ARCHIVE_, _DATA READY_ and _USER_ events without the need to configure polling for the Attribute.

I think we should remove the mention to PIPE events a bit above too. PIPE are not attributes so we should not mention that in this RFC which is dedicated to attributes.

4/README.md
mliszcz: Do you mean that it will be rejected by Pogo or will not work with tools like Jive?
I already added a note:
"Also note that the Attribute name SHOULD contain at least one letter and SHOULD NOT start with a digit."

In my opinion the proper place to put any restrictions on attribute name would be constructor `Attr::Attr`. Otherwise we will have a specification which is not followed by any implementation.

4/README.md
mliszcz: So I've removed the confusing note about the `\0`. I've left the information about multi-byte characters.
* Tango should work (with minor tweaking, see [TangoDatabase#16](https://github.com/tango-controls/TangoDatabase/issues/16)) with any Database chacter set.
* According to the answers in [this stackoverflow question](https://stackoverflow.com/questions/402283/stdwstring-vs-stdstring), `std::string` can hold unicode.
* Simple test shows it is possible to read attribute with non-ASCII name.

mliszcz:
Thanks a lot @bourtemb for an in-depth review and a lot of useful suggestions!

@srgblnch, in the lates commit (87f0d0d) I did a rework according to @bourtemb's suggestions. Please check if you are ok with proposed changes.

These still need to be addressed:
* [x] read / write and other access types in "Goals" section
* [x] unicode names
* [x] *memorized* restrictions and differences between Java and C++
* [x] warning deprecation - I do not remember when I have found this, maybe I'm wrong, need to check
* [x] empty names

4/README.md
bourtemb: > Do you mean that it will be rejected by Pogo or will not work with tools like Jive?

I mean there will be some unexpected side effects for sure. I didn't try it so I don't know how bad it will behave actually.

> I already added a note:
> "Also note that the Attribute name SHOULD contain at least one letter and SHOULD NOT start with a digit."

Indeed, I saw this. The abnf definition is much more relaxed. Maybe we should add a recommended attribute name abnf definition? I understand your point of view. You're simply documenting the current implementation.
I think we should not tolerate empty attribute names in next major Tango versions.

> 
> In my opinion the proper place to put any restrictions on attribute name would be constructor `Attr::Attr`. 

I agree with you

srgblnch:
> > a) In the goals section, there are some more attribute types* than 'read only' and 'read write',
> 
> My intention was to not go into details in the goals section. Shall I reword the goals and include all attribute access types?
> > b) like in the commands rfc, I suggest to reference the Object Oriented paradigm
> 
> Sure, I'll add this information.

I change my mind and I propose that would be better if here, in an introduction level we mention that the concept of a tango attribute comes from the concept of an attribute in object oriented paradigm. This is merging items 'a' and 'b'. Not going much to the detail, as a high light. Later in the 'Attr definition' section is where the current access can be and are described.

about 'c'. Good point the comment about pogo restriction. We may think that pogo is part of the standard so this specification may include it and one day we can move this restriction down (instead of being in the top of our tools).

about 'd'. Yes, it is true that guidelines are subjective and specifications tries to be objective. But when the specification of something that is already implemented supports something that can cause problems, imho is when we can introduce this amendments.

Even if the implementation allows an empty string as attribute name, how we later can reference it because its name collides with the device name. What do you think?

srgblnch:
Hi, @bourtemb 
I've seen you reference an open issue in cppTango about this limitation on memorize writable DevState attributes. What I think:
1) as far as we are describing what we have now a days, I think the specification should say that those attributes cannot be memorized. If java implementation allows it, nice, but don't make the cppTango go below the specs.
2) A clear code in cppTango that avoid explicitly DevStates and DevEncodeds, we may have to think during a while why someone did it. Best case, if how wrote it thought they are not necessary, but perhaps there is a hard reason behind.
Imho, We should describe what we have nowadays and avoid to make our implementations not compilant with the specification. So I prefer to specify the limit and extend it when all our implementations extend it.
What do you think?

bourtemb:
> Hi, @bourtemb
> I've seen you reference an open issue in cppTango about this limitation on memorize writable DevState attributes. What I think:
> 
>     1. as far as we are describing what we have now a days, I think the specification should say that those attributes cannot be memorized.

If you write something like that it means that Java implementation doesn't comply to the specs because it looks like the specs says that it is forbidden or impossible to memorize these kinds of attributes.
Maybe we just need to find the right way to express it. 

>     2. A clear code in cppTango that avoid explicitly DevStates and DevEncodeds, we may have to think during a while why someone did it.

DevEncoded are currently forbidden because this can imply a lot of data because DevEncoded are very often used to store images. So it is currently forbidden for performance reasons.
For the DevState, I guess it is historical but nobody knows why this check is here.

>  Best case, if how wrote it thought they are not necessary, but perhaps there is a hard reason behind.

@taurel does not remember why this check is here for the DevState attributes and doesn't see a good reason to forbid DevState memorized attributes.

>        Imho, We should describe what we have nowadays and avoid to make our implementations not compilant with the specification.

I agree, but we have to take into account the differences between the current implementations (Java, C++,...). I think it is a good exercise to identify these differences too.
The different implementations do not support the features at the same level.
What about the Labview implementation? I don't think we should base our specs on the implementation having the lowest number of features.
I think the main difficulty of this rfc writing exercise is to define precisely what we want to describe.
If we want to describe the current implementation**s**, then, I think maybe we should have different sections for the different implementations because they are different. 
The main concepts are the same, and I think this is what is the most important to be described, but the behaviour can be slightly different between implementations and some features might not be supported in some implementations but in this case we should not write in the specs that these features cannot be implemented or are forbidden.
I'm not against notes specifying the limits of the current implementations though.

> So I prefer to specify the limit and extend it when all our implementations extend it.
>        What do you think?

For me, what is the most important here is the concept of memorized attribute.
I think this feature can be optional in some implementations if performances would be too bad.

mliszcz:
Hi @bourtemb and @srgblnch 

How about rewording:
```
To set memorized, following constraints MUST be satisfied:
    data format MUST be SCALAR,
    data type MUST NOT be STATE or ENCODED,
    writable MUST NOT be READ or READ_WITH_WRITE.
```
Into:
```
A conforming implementation MUST allow to set "memorized" flag and support memorization
for attributes which:
    data format is SCALAR,
    data type is not STATE or ENCODED,
    "writable" flag is not READ or READ_WITH_WRITE.
```
This is true both for Java and C++ implementation. If JTango supports other data formats or if cppTango will decide to lift the constraints for DevState types, they will still be in-line with the spec.

The problem is that we should document the features available in various implementations somewhere. Maybe in the documentation? IMO, RFC is not a good place to mention any particular implementation.

bourtemb:
> Hi @bourtemb and @srgblnch
> 
> How about rewording:
> [...]
> 
> Into:
> 
> ```
> A conforming implementation MUST allow to set "memorized" flag and support memorization
> for attributes which:
>     data format is SCALAR,
>     data type is not STATE or ENCODED,
>     "writable" flag is not READ or READ_WITH_WRITE.
> ```
> 
> This is true both for Java and C++ implementation. If JTango supports other data formats or if cppTango will decide to lift the constraints for DevState types, they will still be in-line with the spec.
> 

Sounds good to me. I would just replace the last sentence with:
"writable" flag is WRITE or READ_WRITE.

> The problem is that we should document the features available in various implementations somewhere. Maybe in the documentation? 

Indeed the doc could be a good place.

>IMO, RFC is not a good place to mention any particular implementation.

In theory it should not indeed...

mliszcz:
Hi @srgblnch and @bourtemb 

I have addressed the remaining points:

* read / write and other access types in "Goals" section
  * no change needed as discussed with Sergi on slack
* unicode names
  * removed, officially we support only ASCII letters and numbers (and this is what is tested)
* memorized restrictions and differences between Java and C++
  * reworded so that implementations can deliver more features than the specification requires.
* warning deprecation
  * I couldn't find the source. Probably this is my mistake. I added a description for min_ and max_warning.

Please take a look if you are ok with recent changes.
Thanks!

srgblnch (approved):

PR #36, opened 2019-07-24
WIP Device Model (RFC-2)

piogor:
The Device Model RFC :

- [x] Goal
- [x] General specification / description
- [x] Use cases
- [x] Device interface
- [x] Naming convetion
- [x] Lifecycle (reference to device server)
- ~~ Document url ~~
- [x] Description
-  ~~black box~~ move to Server RFC
-  ~~reference to events~~ Better that Event references the Device
-  ~~reference to logging~~ Better that Logging references the Device 
-  ~~reference to polling/cache~~ Better that Cache references the Device

### DIFF
--- a/2/README.md
+++ b/2/README.md
@@ -102,7 +102,7 @@ Attribute, Pipe, Command following the rules below:
 
 * The Device SHALL have one unique identifier which represents its Device Name.
 
-###Naming convention
+### Naming convention
 
 * The Device Name SHALL use the following convention:
 
--- a/2/README.md
+++ b/2/README.md
@@ -78,33 +78,29 @@ Typical use cases for Device are:
 ## Specification
 
 A Tango Device is a strict definition of a distributed object. 
-* configuration are represented in the form of properties.
-* data are represented in the form of attributes and pipes.
-* actions are represented in the form of commands.
+* configuration is represented in the form of properties.
+* data are represented in the form of Attributes and Pipes.
+* actions are represented in the form of Commands.
 
 Its model can be represented as a defined tree which each elements are from a defined types: Class, Device, Property, 
 Attribute, Pipe, Command following the rules below:
 
-* The Device is a distributed object which SHALL be accessed locally or via network.
+* The Device is a distributed object which SHALL be accessible locally or via network.
 * The Device SHALL be an instance of one Device Class, see [RFC-9]()
 * The Device MAY have one or several Property, called Device Property, see [RFC-5]()
 * The Device MAY have one or several Attribute, see [RFC-4]() 
 * The Device MAY have one or several Pipe
 * The Device MAY have one or several Command, see [RFC-3]()
 
+* The Device SHALL have one Init command
 
 * The Device SHALL have one State attribute 
 * The Device SHALL have one Status attribute
 * The Device SHALL have one State command
 * The Device SHALL have one Status command
 
+
 **Note**: Future specification may remove 'SHALL have' requirements for State and Status commands.
-
-
-[TO-DO: Should the two points below be moved to Class RFC?]
-* The Device Class MAY have one or more Device instance
-* The Device Class MAY have one or several Property, called Class Property
-
 
 * The Device SHALL have one unique identifier which represents its Name
 
@@ -119,15 +115,32 @@ family = 1*VCHAR
 member = 1*VCHAR
 ```
 
+### Device Class
+
+A Device Class is an association of a list of Device Class Properties, a list of Attributes, a list of Pipes, 
+a list of Commands, a list of Device Properties with a Device Class Name.  
+
+* The Device Class Name SHOULD reflect its instances application context. 
+* The Device Class Name SHALL use the following convention:
+
+``` ABNF
+device-class-name = 1*VCHAR
+```
+
+* The Device Class MAY have one or more Device instance
+* The Device Class MAY have one or several Property, called Class Property
+
+* The Device SHALL expose (provide access to) all Attributes, Pipes and Commands defined by its Device Class.
  
 ### Device Interface
 
-The Device Interface is a list of Attributes, Pipes and Commands provided by a Device Instance.
-The Device instance SHALL expose (provides access to) all Attributes, Pipes and Commands defined by its Device Class.
+The Device Interface is a list of Attributes, Pipes and Commands provided by a Device.
 
-The Device instance MAY expose Attributes and/or Commands not defined by its Device Class. 
-These are called Dynamic Attributes and/or Dynamic Commands respectively, see [RFC-XX]().
-The Dynamic Attributes and Dynamic Commands MAY be added to the interface during Device Initialisation phase and/or Device Operation. 
+* The Device MAY expose Attributes and/or Commands not defined by its Device Class. These are called 
+  Dynamic Attributes and/or Dynamic Commands respectively, see [RFC-XX]().
+
+* The Dynamic Attributes and Dynamic Commands MAY be added to the interface during Device Initialisation phase and/or 
+  Device Operation phase. 
 
 
 ### Device lifecycle
--- a/2/README.md
+++ b/2/README.md
@@ -188,7 +188,7 @@ The Init command purpose is to re-initialise the Device.
 * If not implemented explicitly by the Device Server, State Attribute *read value* SHALL be set to `UNKNOWN`. 
   It is RECOMMENDED to override this behaviour by an adequate implementation of the Device Server.
 
-* If at a given time the Device State is ON and at least one of Device Attribute has the *quality* set to ALARM,
+* If at a given time the Device State is ON and at least one of Device Attributes has the *quality* set to ALARM,
   the Device State *read value* SHALL be set to ALARM. However, this behaviour MAY be overridden by user code 
   in a Device Server implementation.
 
--- a/2/README.md
+++ b/2/README.md
@@ -3,7 +3,7 @@ domain: rfc.tango-controls.org
 shortname: 2/Device
 name: Device Model
 status: raw
-editor: [David Erb](mailto:david.erb@maxiv.lu.se)
+editor: [Vincent Hardion](mailto:vincent.hardion@maxiv.lu.se))
 contributors: [Piotr Goryl](mailto:piotr.goryl@s2innovation.com)
 ---
 
--- a/2/README.md
+++ b/2/README.md
@@ -103,7 +103,7 @@ Attribute, Pipe, Command following the rules below:
 
 **Note**: Future specification may remove 'SHALL have' requirements for State and Status commands.
 
-* The Device SHALL have one unique identifier which represents its Device Name
+* The Device SHALL have one unique identifier which represents its Device Name.
 
 ### Naming convention
 
@@ -157,28 +157,42 @@ The device lifecycle MUST be implemented by a Device Server.
 
 ### Device State and Status
 
-A value of State attribute SHOULD reflect context of a running Device. If Device is related to a hardware or virtual
-entity or to some process, the value of State attribute SHOULD reflect the state of entity or process.   
+* The *data type* of State Attribute MUST be `DevState`. The *read value* of Status attribute MUST be 
+  of `DevState` data type.
+* The State attribute's *writable* metadata MUST be `READ`.
+* The *data format* of State Attribute MUST be `SCALAR`.
+* The *read value* of State Attribute SHOULD reflect the context of a running Device. If Device is related to hardware or a virtual entity or some process, the value of State attribute SHOULD reflect the state of the entity or process.
 
-[TO-DO: default Status attribute value ]
+* The *data type* of Status attribute MUST `DevString`. The *read value* of Status attribute MUST be 
+  of `DevString` data type.
+* The Status attribute's `writable` metadata MUST be `READ`. 
+* The `data format` of the Status attribute MUST be `SCALAR`.
 
-[TO-DO: default state and status command ]
+* A *read value* of Status Attribute SHOULD provide status information related to the Device. 
+* Default implementation SHOULD set Status Attribute's  *read value* to indicate the Device State. 
+* The Status default behaviour MAY be re-implemented with a user-code in Device Server to provide 
+  more or other information. 
+
+* The State Command MUST return the same value as would be read from the State Attribute at the time 
+  of the command execution.
+* The Status Command MUST return the same value as would be read from the Status Attribute at the time of the command execution.
 
 ### State Machine
 
-If not implemented explicitly by the Device Server,  State attribute value is set by default to `UNKNOWN`.
-It is RECOMMENDED to override this behaviour by adequate implementation of the Device Server.
+* If not implemented explicitly by the Device Server, State Attribute *read value* SHALL be set to `UNKNOWN`. 
+  It is RECOMMENDED to override this behaviour by an adequate implementation of the Device Server.
 
-[TO-DO: automatic ALARM state]
+* If at least one of Device Attribute has the *quality* set to `ALARM` the *read value* of Device State attribute 
+  MUST be `ALARM`.
 
-Value of Device's State attribute MAY impact how the Device object respond to:
+Value of Device's State attribute MAY impact how the Device object responds to:
 
 * Writing to a Device's Attribute or to a Device's Pipe  
 * Calling a Device's Command
 
-Device MAY not execute (block) the attribute write or the pipe write operation or the command execution code for certain States.
-In such a case the the Device object SHALL respond with throwing an exception.  [TO-DO: specify which exception]
-The way how a Device object respond to the above calls define a Device State Machine.
+The Device MAY not execute (block) the attribute write or the pipe write operation or the command execution code for individual States.
+In such a case the Device object SHALL respond with throwing a `DevFailed` exception.
+The way how a Device object responds to the above calls define a Device State Machine.
 
 **NOTE**: Current implementation allows a Device object to block the operations irrespectively to the Device's State
 attribute value. 
--- a/2/README.md
+++ b/2/README.md
@@ -152,6 +152,14 @@ The Device lifecycle is:
 
 The device lifecycle MUST be implemented by a Device Server.
 
+### Init command
+
+The Init command purpose is to re-initialise the Device.
+
+* The Init command, when called, SHALL in a sequence:
+  * gracefully un-initialise the Device (free any dynamic resources allocated to the Device)
+  * initialise the Device
+
 ### Device State and Status
 
 * The *data type* of State Attribute MUST be `DevState`. The *read value* of Status attribute MUST be 
--- a/2/README.md
+++ b/2/README.md
@@ -128,7 +128,7 @@ device-class-name = 1*VCHAR
 * The Device Class MAY have one or more Device instance.
 * The Device Class MAY have one or several Property, called Class Property.
 
-* The Device SHALL expose (provide access to) all Attributes, Pipes and Commands defined by its Device Class.
+* The Device SHALL have all Attributes, Pipes and Commands defined by its Device Class.
  
 ### Device Interface
 
--- a/2/README.md
+++ b/2/README.md
@@ -30,7 +30,7 @@ and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119]
 
 ## Tango Device Specification
 
-A Device is designed to represent any controlled object in Tango Controls system.
+A Device is designed to represent any controlled object in Tango Controls System.
 This specification is inspired by the comment written by A. Götz and E. Taurel in the tango.idl file from 
 the original implementation (https://github.com/tango-controls/tango-idl)
 
@@ -39,7 +39,7 @@ the original implementation (https://github.com/tango-controls/tango-idl)
 ### Goals
 
 The Tango Device represents the fundamental interface for all TANGO objects. Directly inspired from the object 
-oriented programming a Device object has data and actions. It allows users of the Tango Controls system to access 
+oriented programming a Device object has data and actions. It allows users of the Tango Controls System to access 
 (read and/or write) information (data) stored in or processed by hardware or virtual devices and call actions which
 may influence hardware state and behaviour.  Typically, access to data is provided via Device's Attributes and Pipes
 and actions are initiated by calling Device's Commands.  
@@ -53,7 +53,7 @@ Additionally, it aims to:
 * Be usable as a logical abstraction of hardware. For example the same GUI application may be used to control power 
   supplies powering magnets and powering heat coils.
 
-* Expose, to Users and other Actors of the Tango Controls system, a semantic interface which is well understood, 
+* Expose, to Users and other Actors of the Tango Controls System, a semantic interface which is well understood, 
   intuitive, and match the object-oriented paradigm. For example, a power supply Device may be implemented in Java, 
   Python or C++ language. Irrespectively to the implementation language, Attributes *current* and *voltage* can be 
   implemented as field members of an object, while *on* and *off* commands can be implemented as methods of the 
@@ -147,10 +147,10 @@ The Device Interface is a list of Attributes, Pipes and Commands provided by a D
 ### Device lifecycle
 
 The Device lifecycle is:
-* Device Initialisation phase, when Device object is created and made visible in a Tango Controls system (see [RFC-1](), [RFC-8]())
-* Device Operation phase, when Device Interface MUST be available to actors of the Tango Controls system.
+* Device Initialisation phase, when Device object is created and made visible in a Tango Controls System (see [RFC-1](), [RFC-8]())
+* Device Operation phase, when Device Interface MUST be available to actors of the Tango Controls System.
 * Device Destruction phase, when Device SHOULD be unregistered from the Tango Control system and Device Interface MAY not
-  be available within the Tango Controls system. 
+  be available within the Tango Controls System. 
 
 The device lifecycle MUST be implemented by a Device Server.
 
--- a/2/README.md
+++ b/2/README.md
@@ -84,7 +84,7 @@ Its model can be represented as a defined tree which each element are from a def
 Attribute, Pipe, Command following the rules below:
 
 * The Device is a distributed object which SHALL be accessible locally or via the network.
-* The Device SHALL be an instance of one Device Class, see [Device Class section](#device-class).
+* The Device SHALL be an instance of one Class, see [Device Class section](#device-class).
 * The Device MAY have one or several Property, called Device Property, see [RFC 5/Property](/5).
 * The Device MAY have one or several Attribute, see [RFC 4/Attribute](/4).
 * The Device MAY have one or several Pipe, see [RFC 7/Pipe](/7).
@@ -115,26 +115,26 @@ member = 1*VCHAR
 
 ### Device Class
 
-A Device Class is an association of a list of Device Class Properties, a list of Attributes, a list of Pipes, 
+A Class is an association of a list of Class Properties, a list of Attributes, a list of Pipes, 
 a list of Commands, a list of Device Properties with a Device Class Name.  
 
-* The Device Class Name SHOULD reflect its instances application context. 
-* The Device Class Name SHALL use the following convention:
+* The Class Name SHOULD reflect its instances application context. 
+* The Class Name SHALL use the following convention:
 
 ``` ABNF
-device-class-name = 1*VCHAR
+class-name = 1*VCHAR
 ```
 
-* The Device Class MAY have one or more Device instance.
-* The Device Class MAY have one or several Property, called Class Property.
+* The Class MAY have one or more Device instance.
+* The Class MAY have one or several Property, called Class Property.
 
-* The Device SHALL have all Attributes, Pipes and Commands defined by its Device Class.
+* The Device SHALL have all Attributes, Pipes and Commands defined by its Class.
  
 ### Device Interface
 
 The Device Interface is a list of Attributes, Pipes and Commands provided by a Device.
 
-* The Device MAY expose Attributes and/or Commands not defined by its Device Class. These are called 
+* The Device MAY expose Attributes and/or Commands not defined by its Class. These are called 
   Dynamic Attributes and/or Dynamic Commands respectively, see [RFC-XX]().
 
 * The Dynamic Attributes and Dynamic Commands MAY be added to the interface during Device Initialisation phase and/or 
@@ -217,10 +217,10 @@ member = 1*VCHAR
 
 `admin-device-name` MUST not be used for Devices of other than `DServer` Device Class.  
 
-### Reserved Device Class Names
+### Reserved Class Names
 
-Devices of some of Device Classes are used to provide standard Tango Controls System services. 
-These Device Classes' Names SHOULD not be used for other purposes. Below is a list of reserved Device Class Names:
+Devices of some of Classes are used to provide standard Tango Controls System services. 
+These Classes' Names SHOULD not be used for other purposes. Below is a list of reserved Class Names:
 
 * `%i"DataBaseds"`, see [RFC 6/Database](/6)
 * `%i"TangoAccessControl"`
--- a/2/README.md
+++ b/2/README.md
@@ -39,7 +39,7 @@ the original implementation (https://github.com/tango-controls/tango-idl)
 
 The Tango Device represents the fundamental interface for all TANGO objects. Directly inspired by object-oriented programming, a Device object has data and actions. It allows users of the Tango Controls System to access 
 (read and/or write) information (data) stored in or processed by hardware or virtual devices and call actions which
-may influence hardware state and behaviour.  Typically, access to data is provided via Device's Attributes, and Pipes
+may influence hardware state and behaviour.  Typically, access to data is provided via Device's Attributes and Pipes
 and actions are initiated by calling Device's Commands.  
 
 Additionally, it aims to:
@@ -90,15 +90,15 @@ Attribute, Pipe, Command following the rules below:
 * The Device MAY have one or several Pipe, see [RFC 7/Pipe](/7).
 * The Device MAY have one or several Command, see [RFC 3/Command](/3).
 
-* The Device SHALL have one Init command.
+* The Device SHALL have one Init Command.
 
-* The Device SHALL have one State attribute. 
-* The Device SHALL have one Status attribute.
-* The Device SHALL have one State command.
-* The Device SHALL have one Status command.
+* The Device SHALL have one State Attribute. 
+* The Device SHALL have one Status Attribute.
+* The Device SHALL have one State Command.
+* The Device SHALL have one Status Command.
 
 
-**Note**: Future specification may remove 'SHALL have' requirements for State and Status commands.
+**Note**: Future specification may remove 'SHALL have' requirements for State and Status Commands.
 
 * The Device SHALL have one unique identifier which represents its Device Name.
 
@@ -156,11 +156,11 @@ The Device lifecycle is:
 
 The device lifecycle MUST be implemented by a Device Server.
 
-### Init command
+### Init Command
 
-The Init command purpose is to re-initialise the Device.
+The Init Command purpose is to re-initialise the Device.
 
-* The Init command, when called, SHALL in a sequence:
+* The Init Command, when called, SHALL in a sequence:
   * gracefully un-initialise the Device (free any dynamic resources allocated to the Device)
   * initialise the Device
 
@@ -168,17 +168,17 @@ The Init command purpose is to re-initialise the Device.
 
 The State Attribute represents Device State. 
 
-* The *data type* of State Attribute MUST be `DevState`. The *read value* of Status attribute MUST be 
+* The *data type* of State Attribute MUST be `DevState`. The *read value* of Status Attribute MUST be 
   of `DevState` data type.
-* The State attribute's *writable* metadata MUST be `READ`.
+* The State Attribute's *writable* metadata MUST be `READ`.
 * The *data format* of State Attribute MUST be `SCALAR`.
 * The *read value* of State Attribute SHOULD reflect the context of a running Device. If Device is related to hardware
-  or a virtual entity or some process, the value of State attribute SHOULD reflect the state of the entity or process.
+  or a virtual entity or some process, the value of State Attribute SHOULD reflect the state of the entity or process.
 
-* The *data type* of Status attribute MUST `DevString`. The *read value* of Status attribute MUST be 
+* The *data type* of Status Attribute MUST `DevString`. The *read value* of Status Attribute MUST be 
   of `DevString` data type.
-* The Status attribute's `writable` metadata MUST be `READ`. 
-* The `data format` of the Status attribute MUST be `SCALAR`.
+* The Status Attribute's `writable` metadata MUST be `READ`. 
+* The `data format` of the Status Attribute MUST be `SCALAR`.
 
 * A *read value* of Status Attribute SHOULD provide status information related to the Device. 
 * Default implementation SHOULD set Status Attribute's  *read value* to indicate the Device State. 
@@ -198,7 +198,7 @@ The State Attribute represents Device State.
   the Device State *read value* SHALL be set to ALARM. However, this behaviour MAY be overridden by user code 
   in a Device Server implementation.
 
-Value of Device's State attribute MAY impact how the Device object responds to:
+Value of Device's State Attribute MAY impact how the Device object responds to:
 
 * Reading and writing of its Attribute or Pipe  
 * Calling a Command
--- a/2/README.md
+++ b/2/README.md
@@ -198,10 +198,9 @@ Value of Device's State attribute MAY impact how the Device object responds to:
 
 The Device MAY not execute (block) the Attribute or Pipe read or write operation or the command execution code 
 for individual States. In such a case the Device object SHALL respond with throwing a `DevFailed` exception.
-The way how a Device object responds to the above calls define a Device State Machine.
+The way how a Device object responds to the above calls define a State Machine.
 
-**NOTE**: Current implementation allows a Device object to block the operations irrespectively to the Device's State
-attribute value. 
+* A Device MAY use a Stat Machine.
 
 ### Reserved Device Names
 
--- a/2/README.md
+++ b/2/README.md
@@ -195,8 +195,8 @@ Value of Device's State Attribute MAY impact how the Device responds to:
 * Calling a Command.
 
 The Device MAY not execute (block) the Attribute or Pipe read or write operation or the command execution code 
-for individual States. In such a case the Device object SHALL respond with throwing a `DevFailed` exception.
-The way how a Device object responds to the above calls define a State Machine.
+for individual States. In such a case the Device SHALL respond with throwing a `DevFailed` exception.
+The way how a Device responds to the above calls define a State Machine.
 
 * A Device MAY use a State Machine.
 
--- a/2/README.md
+++ b/2/README.md
@@ -90,7 +90,7 @@ device-name = domain "/" family "/" member
 domain = 1*VCHAR
 family = 1*VCHAR
 member = 1*VCHAR
-``
+```
 
 ### Device lifecycle
 
--- a/2/README.md
+++ b/2/README.md
@@ -42,11 +42,18 @@ Additionally, it aims to:
 
 ### Use Cases
 
-There are X main use cases for Template:
+Typical use cases for Device are:
 
-* To reduce the entry cost of RFC submission.
+* It is common for particle accelerator systems to use a set of power supplies for powering a magnet system. 
+  Power supplies may come from a different vendors. These power supplies may use different communication interfaces
+  and protocols for remote control. However, there is a need to steer these power supplies from a remote locations in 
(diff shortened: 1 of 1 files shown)


2/README.md
```diff
@@ -54,29 +54,62 @@ There are X main use cases for Template:
...
 Its model can be represented as a defined tree which each elements are from a defined types: Class, Device, Property, Attribute, Command following the rules below:
 
 
 * The Device is a distributed object which SHALL be accessed locally or via network.
-* The Device SHALL be an instance of one Class
+* The Device SHALL be an instance of one Device Class (see RFC-9)
 * The Device MAY have one or several Property, called Device Property
-* The Device MAY have one or several Attribute
+* The Device MAY have one or several Attribute, 
 * The Device SHALL have one State attribute
```
bourtemb: This seems to be missing:
The device SHALL have one State command
The device SHALL have one Status command

lorenzopivetta: As far as I remember, State and Status Commands are sort of legacy stuff, before they got properly implemented as Attributes. Indeed, State and Status Commands are in the current model, thus showing up in Tango 9 RFC, but I'm not sure they need to be there for Tango 10. Your thoughts? Does it make sense to capture in a note?

bourtemb: > Indeed, State and Status Commands are in the current model, thus showing up in Tango 9 RFC, but I'm not sure they need to be there for Tango 10.

My comment was indeed only for Tango 9 RFC. I think we should get rid of State and Status mandatory commands in "Tango 10".

piogor: Thanks, @bourtemb and @lorenzopivetta. The below lines add the missing commands and note on future plans.

2/README.md
```diff
@@ -54,29 +54,104 @@ There are X main use cases for Template:
...
+* The Device SHALL have one State command
+* The Device SHALL have one Status command
+
+**Note**: Future specification may remove 'SHALL have' requirements for State and Status commands.
+
+
+[TO-DO: Should the two points below be moved to Class RFC?]
+* The Device Class MAY have one or more Device instance
+* The Device Class MAY have one or several Property, called Class Property
+
```
piogor, line 102: @hardion, the points above seem to belong to Device Class (RFC-9). Since there is no RFC-9 yet, should I create one and move these there, so it will not be forgotten?

2/README.md
```diff
@@ -54,29 +54,104 @@ There are X main use cases for Template:
...
+The Device Server MUST implement at least Object creation step.
+
+#### Device Operation
+
+* Attribute read and write
+* Command Call
+* Device State handling
+
+#### Device Destruction
+
```
piogor, line 221: @hardion, there is a similar question as for Device Class RFC. Should I create RFC-8 and move the text there?

david-erb:
Will events be considered within RFC-2?

david-erb:
Same question about black box, is that concept meant to be part of the Tango Device Model?

piogor:
@david-erb-maxiv 
> Will events be considered within RFC-2?

I think these should be mentioned (I will do it) but there is a separate RFC (12).

piogor:
@david-erb-maxiv 
> Same question about black box, is that concept meant to be part of the Tango Device Model?

This probably belongs to the device model. I will add it.

2/README.md
```diff
@@ -54,29 +54,62 @@ There are X main use cases for Template:
 
 A Tango Device is a strict definition of a distributed object. 
 * configuration are represented in the form of properties.
-* data are represented in the form of attributes.
+* data are represented in the form of attributes and pipes.
```
hardion: Pipes are attributes, aren't they?

2/README.md
hardion: It’s probably better to start writing the others RFCs but if there is not much to say ,i.e a Class, maybe it can stay in the Device RFC with a particular chapter.

2/README.md
hardion: You meant Device Server. There is probably more to say about the Device Server since the all interface has to be defined.
You can simply add a Reserved Devices chapter where the domain "dserver" is reserved and point to the RFC/8.

2/README.md
hardion: @lorenzopivetta @bourtemb 
Can you make an issue for the commands to remove with the tag v10?

2/README.md
```diff
@@ -3,7 +3,7 @@ domain: rfc.tango-controls.org
 shortname: 2/Device
 name: Device Model
 status: raw
-editor: Vincent Hardion (vincent.hardion@maxiv.lu.se)
+editor: David Erb (david.erb@maxiv.lu.se)
 ---
```
hardion, line 8: Missing the contributors

2/README.md
```diff
@@ -3,7 +3,7 @@ domain: rfc.tango-controls.org
...
 
 This document describes the Tango Device model specification version 5.0.
```
hardion, line 10: Below the references:
- need the RFC number
- add Class
- add Tango Control System (Mention in Tango Device Specification)

2/README.md
```diff
@@ -22,15 +22,15 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 
 ## Tango Device Specification
 
-A Device is designed to represent any controlled object in Tango.
+A Device is designed to represent any controlled object in Tango Controls system.
```
hardion: Tango Controls system => Tango Control System

bourtemb:
>  @lorenzopivetta @bourtemb
Can you make an issue for the commands to remove with the tag v10?

Done: See #37

2/README.md
```diff
@@ -54,29 +54,104 @@ There are X main use cases for Template:
 
 A Tango Device is a strict definition of a distributed object. 
 * configuration are represented in the form of properties.
-* data are represented in the form of attributes.
+* data are represented in the form of attributes and pipes.
```
hardion: Here the word attributes does not refer to the Tango Attributes but the attributes in object programming. Pipe does not exist in Object Programming.

2/README.md
```diff
@@ -54,29 +54,104 @@ There are X main use cases for Template:
...
-Its model can be represented as a defined tree which each elements are from a defined types: Class, Device, Property, Attribute, Command following the rules below:
-
+Its model can be represented as a defined tree which each elements are from a defined types: Class, Device, Property, 
+Attribute, Pipe, Command following the rules below:
 
 * The Device is a distributed object which SHALL be accessed locally or via network.
-* The Device SHALL be an instance of one Class
-* The Device MAY have one or several Property, called Device Property
-* The Device MAY have one or several Attribute
+* The Device SHALL be an instance of one Device Class, see [RFC-9]()
```
hardion: I guess you replace "Class" by "Device Class" to be more precise but does it exist another type of Class? If not I would keep the simple term Class.

2/README.md
```diff
@@ -54,29 +54,104 @@ There are X main use cases for Template:
...
+
+
 * The Device SHALL have one State attribute 
 * The Device SHALL have one Status attribute
-* The Device MAY have one or several Command
-* The Device SHALL have one unique identifier which represents its Name
+* The Device SHALL have one State command
+* The Device SHALL have one Status command
+
+**Note**: Future specification may remove 'SHALL have' requirements for State and Status commands.
```
hardion: Remove the note as it's not decided yet.
On the other hand you can specify:
- what is the purpose of State and Status,
- that State Command MUST return the same value as the State Attribute 
- (same for Status)
- that State Attribute is the only reference for the State Machine

2/README.md
```diff
@@ -54,29 +54,104 @@ There are X main use cases for Template:
...
+
+
+[TO-DO: Should the two points below be moved to Class RFC?]
+* The Device Class MAY have one or more Device instance
+* The Device Class MAY have one or several Property, called Class Property
+
 
-* The Class MAY have one or more Device instance
-* The Class MAY have one or several Property, called Class Property
+* The Device SHALL have one unique identifier which represents its Name
```
hardion: its Name => the Device Name

2/README.md
```diff
@@ -54,29 +54,104 @@ There are X main use cases for Template:
...
 
 ### Naming convention
+
 * The Device's Name SHALL use the following convention:
+
```
hardion, line 108: Missing the Device Alias

2/README.md
```diff
@@ -54,29 +54,104 @@ There are X main use cases for Template:
...
 device-name = domain "/" family "/" member
 domain = 1*VCHAR
 family = 1*VCHAR
 member = 1*VCHAR
 ```
+
+ 
+### Device Interface
+
+The Device Interface is a list of Attributes, Pipes and Commands provided by a Device Instance.
```
hardion: What is the difference between "Device" and "Device Instance"?
Also I would define new key word i.e "Device Interface" only if:
- commonly used in the Tango Control System
- the term is helping the syntax in the specification. Here it's only use 2 times later in the Device lifecycle paragraph

2/README.md
```diff
@@ -54,29 +54,104 @@ There are X main use cases for Template:
...
+The Device instance SHALL expose (provides access to) all Attributes, Pipes and Commands defined by its Device Class.
```
hardion: Here the reader can be confused with the object paradigm where class attributes are static and shared between its instances, which is not the case in Tango. 
Can you rephrase  the statement that Device SHALL defined the same Attribute, ... than its Class.

2/README.md
hardion: Also Class would define a Device Interface as well

2/README.md
```diff
@@ -54,29 +54,104 @@ There are X main use cases for Template:
...
+
+The Device instance MAY expose Attributes and/or Commands not defined by its Device Class. 
+These are called Dynamic Attributes and/or Dynamic Commands respectively, see [RFC-XX]().
+The Dynamic Attributes and Dynamic Commands MAY be added to the interface during Device Initialisation phase and/or Device Operation. 
+
```
hardion, line 140: Missing the Init Command and the purpose/use case

2/README.md
```diff
@@ -54,29 +54,104 @@ There are X main use cases for Template:
...
+
+### Device lifecycle
```
hardion, line 143: I am bit perplex of this paragraph Device lifecycle: It makes sense to have a notion of life cycle for device. But to reuse the wording in Tango, a device has only two phases: 
- A Device is said Exported when this Device MUST be accessible ...
- A Device is said Not Exported when this Device MUST NOT be accessible .... 

- At the transition from the phase Not Exported to the phase Exported the Device MUST be initialised as if its Init Command was executed. (I don't want to say the Device must execute the Init Command because the current implementation)
- At the transition from the phase Exported to the phase Not Exported the Device SHOULD gracefully be uninitialised. (There is no Command Delete - Only the Device Server has a command Kill )

2/README.md
```diff
@@ -54,29 +54,104 @@ There are X main use cases for Template:
...
+
+The Device lifecycle is:
+* Device Initialisation phase, when Device object is created and made visible in a Tango Controls system (see [RFC-1](), [RFC-8]())
+* Device Operation phase, when Device Interface MUST be available to actors of the Tango Controls system.
+* Device Destruction phase, when Device SHOULD be unregistered from the Tango Control system and Device Interface MAY not
```
hardion: A device is not unregistered when not not exported

2/README.md
```diff
@@ -54,29 +54,104 @@ There are X main use cases for Template:
...
+  be available within the Tango Controls system. 
+
+The device lifecycle MUST be implemented by a Device Server.
```
hardion: To move to Device Server

2/README.md
```diff
@@ -54,29 +54,104 @@ There are X main use cases for Template:
...
+
+### Device State
```
hardion: Missing at his point the purpose/use case of the State Attribute i.e reflect of the hardware connection, ...
After this state is used for the State Machine but this is not the primary use case.

hardion: Missing the default value (INIT) of the State when the device got exported.

2/README.md
```diff
@@ -54,29 +54,104 @@ There are X main use cases for Template:
...
+
+Value of Device's State attribute MAY impact how the Device object respond to:
```
hardion: Maybe a paragraph called "State Machine" is suitable here

2/README.md
```diff
@@ -54,29 +54,104 @@ There are X main use cases for Template:
...
+
+* Writing to a Device's Attribute or to a Device's Pipe
```
hardion: Reading as well.
Minor comment: you can remove "Device's". It's implicit as Attribute, Pipe, Command exists only in the context of a device

2/README.md
```diff
@@ -54,29 +54,104 @@ There are X main use cases for Template:
...
+The device lifecycle MUST be implemented by a Device Server.
+
+### Device State
+
+Value of Device's State attribute MAY impact how the Device object respond to:
+
+* Writing to a Device's Attribute or to a Device's Pipe  
+* Calling a Device's Command
+
+Device MAY not execute (block) the attribute write or the pipe write operation or the command execution code for certain States.
```
hardion: Sounds like a repetition of the previous paragraph "Value of Device's State...".
Maybe you should merge them and use the word "allowed" (Device which execute an attribute writing may be confusing some the reader)

2/README.md
```diff
@@ -54,29 +54,104 @@ There are X main use cases for Template:
...
+In such a case the the Device object SHALL respond with throwing an exception.  [TO-DO: specify which exception]
+The way how a Device object respond to the above calls define a Device State Machine.
+
+**NOTE**: Current implementation allows a Device object to block the operations irrespectively to the Device's State
```
hardion: So it has to be clear if having a State Machine is mandatory or not.

2/README.md
```diff
@@ -54,29 +54,104 @@ There are X main use cases for Template:
...
+
+
+[TO-DO: Move the following to Device Server ? ]
+#### Device Initialization
+
+Device Initialisation phase is intended to prepare Device Object for Operation. 
+
+Device Initialization SHOULD be divided into two subsequent steps:
+
+1) Object creation, when an object is created in a space of a Device Server process. After this phase Device object
```
hardion: A quick comment even if you move the text.
- Nothing prevents today to have the server and the device in different process. This is an implementation details.
- The term Object is not defined.

2/README.md
```diff
@@ -54,29 +54,104 @@ There are X main use cases for Template:
...
+
+[TO-DO: Move the following to Device Server ? ]
+#### Device Initialization
+
+Device Initialisation phase is intended to prepare Device Object for Operation. 
+
+Device Initialization SHOULD be divided into two subsequent steps:
+
+1) Object creation, when an object is created in a space of a Device Server process. After this phase Device object 
+   SHALL be accessible locally and via network. During this phase the same `user code` which is executed when the
```
hardion: The same 'user code' is an implementation details. I think even this is not the case of libtango where the Init Command is executing delete_device() AND init_device(), while only init_device() is called at the start up.

2/README.md
```diff
@@ -54,29 +54,104 @@ There are X main use cases for Template:
...
+
+Device Initialisation phase is intended to prepare Device Object for Operation. 
+
+Device Initialization SHOULD be divided into two subsequent steps:
+
+1) Object creation, when an object is created in a space of a Device Server process. After this phase Device object 
+   SHALL be accessible locally and via network. During this phase the same `user code` which is executed when the 
+   Device Init command is executed MAY be called.
+
+2) Device exporting, when Device is made visible within the Tango Controls system. It SHALL use a Database System
```
hardion: Shall use a Database System => File can be used as well (but Server RFC should may or not care of these details)

2/README.md
```diff
@@ -54,29 +54,104 @@ There are X main use cases for Template:
...
+
+2) Device exporting, when Device is made visible within the Tango Controls system. It SHALL use a Database System 
+   ([RFC-6]()) to provide information on how other actors in the system can access its interface. 
+
+
+The Device Server MUST implement at least Object creation step.
+
+#### Device Operation
+
+* Attribute read and write
```
hardion: One idea for this sections: Allow multiple read/write of Attributes

hardion (commented):
> @david-erb-maxiv
> 
> > Will events be considered within RFC-2?
> 
> I think these should be mentioned (I will do it) but there is a separate RFC (12).

The RFC/12 Publish Subscriber will reference this RFC. No need to mention here.

gwen-soleil:
Hi,

I think that some operations defined in the IDL (https://github.com/tango-controls/tango-idl/blob/tango-9-lts/tango.idl) are missing in this specification:
* info 
* ping
* adm_name
* name
* description
* get_attribute_config and set_attribute_config
* read_attribute_history
* command_list_query

hardion:
Thanks for the review. The specification are at a higher level than the IDL so some information are dispatched among the RFCs. See below my comment

> I think that some operations defined in the IDL (https://github.com/tango-controls/tango-idl/blob/tango-9-lts/tango.idl) are missing in this specification:
> * info
DevInfo is a structure of different information related to a device but not necessary of the device.
```
	string 	dev_class;
```
Line 85: The Device SHALL be an instance of one Device Class, see RFC-9

```
	string 	server_id;
	string 	server_host;
	long 	server_version;
```
Good catch! Should it be in the RFC Server ?

```
	string 	doc_url;
```
Sounds something to add to the Device 

```
	string 	dev_type;
```
Do you know what is this parameter?

> * ping
It sounds more part of the Reply Request protocol.

> * adm_name
It can be define in the Server RFC

> * name
line 105: The Device SHALL have one unique identifier which represents its Name

> * description
Yes, it should be part of this RFC

> * get_attribute_config and set_attribute_config
Should be part of the Attribute RFC

> * command_list_query
line 93: The Device MAY have one or several Command

2/README.md
hardion: @piogor Do you think you can add this reserved domain "dserver" to the RFC this week?
You can take example to https://github.com/tango-controls/rfc/blob/draft/3/README.md#reserved-commands

hardion:
> Will events be considered within RFC-2?

Event will be specified in the RFC 10

hardion:
> Same question about black box, is that concept meant to be part of the Tango Device Model?

Will be specified in the Server model RFC8 (Can be part of the request reply)

2/README.md
hardion: @piogor Can you add the contributors this week?
```md
contributor: Piotr Goryl (piotr.goryl@s2innovation.com)
```

2/README.md
hardion:
```md
See also: 1/ControlSystem 3/Command 4/Attribute 5/Property X/Pipe 
```
Maybe pipe can be RFC 7

2/README.md
piogor: Added, should I also update the editor to point to you @hardion?

2/README.md
piogor: Thanks, updated.

2/README.md
piogor: @hardion,  someone may reference Class to a general concept of Class in OOP. I would stick to Device Class.

2/README.md
```diff
@@ -3,80 +3,178 @@ domain: rfc.tango-controls.org
...
+* configuration is represented in the form of properties.
+* data are represented in the form of Attributes and Pipes.
+* actions are represented in the form of Commands.
+
+Its model can be represented as a defined tree which each elements are from a defined types: Class, Device, Property, 
+Attribute, Pipe, Command following the rules below:
 
-Its model can be represented as a defined tree which each elements are from a defined types: Class, Device, Property, Attribute, Command following the rules below:
+* The Device is a distributed object which SHALL be accessible locally or via network.
+* The Device SHALL be an instance of one Device Class, see [RFC-9]()
```
piogor: @hardion, probably we should remove references to RFC-9, as all about Device Class is in this document. What do you think?

2/README.md
lorenzopivetta: Somehow difficult to read... what about adding the "exported" concept to the Device Initialization step?

2/README.md
```diff
@@ -3,80 +3,178 @@ domain: rfc.tango-controls.org
...
+  Dynamic Attributes and/or Dynamic Commands respectively, see [RFC-XX]().
+
+* The Dynamic Attributes and Dynamic Commands MAY be added to the interface during Device Initialisation phase and/or 
+  Device Operation phase. 
+
+
+### Device lifecycle
+
+The Device lifecycle is:
+* Device Initialisation phase, when Device object is created and made visible in a Tango Controls System (see [RFC-1](), [RFC-8]())
```
lorenzopivetta: Device Initialisation phase, when Device object is created and made visible (exported) in a Tango...

2/README.md
piogor: @hardion, you have got the point. I will commit this paragraph a bit rephrased.

2/README.md
piogor: Thanks, @lorenzopivetta, I will rephrase the paragraph to use wording Exported and Not Exported.

2/README.md
piogor: I will check if it is not UNKNOWN until someone explicitly calls set_state().

2/README.md
piogor: Finally, I removed references to the RFC-9.

2/README.md
piogor: added

2/README.md
piogor: I think it is worth to keep this here.

2/README.md
piogor: Added as a short paragraph after 'lifecycle'.

2/README.md
piogor: Implemented in the related section.

2/README.md
piogor: Done

2/README.md
piogor: rephrased

2/README.md
piogor: done

2/README.md
piogor: You are right. Corrected

2/README.md
piogor: I have removed the section.

2/README.md
piogor: Rewritten.

2/README.md
piogor: Rewritten.

2/README.md
piogor: Rewritten.

2/README.md
piogor: I think it is worth to keep the section, as the term `state machine` is appearing in the documentation and tools as well as is reflected in the API and generated code.

2/README.md
piogor: Yes, but we are talking about  Device which has this feature.

2/README.md
piogor: Rephrased with the most recent commit.

2/README.md
hardion: Yes please. Other wise I will do it during the merge.

2/README.md
hardion: OK so please use Capital letter for Property, Attribute, Pipe

2/README.md
hardion: @piogor Previously you wrote the same comment for attribute which should be Attribute but the initial idea was to talk about attributes as in OOP. 
The RFC use this convention for the term defined in the text. If one want to talk about the class concept just write with lower case.
An alternative would be to use this "Device X" convention everywhere e.g Device Attribute, Device Command, ...

2/README.md
hardion: @piogor my comment was about the ***NOTE*** of the current implementation.
If the implementation allows to override the State Machine I would rather mention:
```
A Device MAY use a State Machine
```

If you (after asking the kernel contributors) consider it's  a bug:
```
A Device SHALL use a State Machine
```

and remove the NOTE

2/README.md
```diff
@@ -29,18 +31,15 @@ and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119]
...
-### Related Specification
-
 ### Goals
 
-The Tango Device represents the fundamental interface for all TANGO objects. Directly inspired from the object 
-oriented programming a Device object has data and actions. It allows users of the Tango Controls system to access 
+The Tango Device represents the fundamental interface for all TANGO objects. Directly inspired by object-oriented programming, a Device object has data and actions. It allows users of the Tango Controls System to access 
 (read and/or write) information (data) stored in or processed by hardware or virtual devices and call actions which
-may influence hardware state and behaviour.  Typically, access to data is provided via Device's Attributes and Pipes
+may influence hardware state and behaviour.  Typically, access to data is provided via Device's Attributes, and Pipes
```
hardion: Small details the comma means you separate Device's Attribute  from Pipes.
Device's Attributes and Pipes (without comma) is equivalent to Device's Attributes and Device's Pipes

2/README.md
```diff
@@ -70,43 +68,43 @@ Typical use cases for Device are:
...
+* The Device SHALL have one Init command.
 
-* The Device SHALL have one State attribute 
-* The Device SHALL have one Status attribute
-* The Device SHALL have one State command
-* The Device SHALL have one Status command
+* The Device SHALL have one State attribute. 
+* The Device SHALL have one Status attribute.
+* The Device SHALL have one State command.
+* The Device SHALL have one Status command.
```
hardion: I did not catch these early on. 
Init ***C***ommand,
State ***A***ttribute,
...

2/README.md
```diff
@@ -127,10 +125,10 @@ a list of Commands, a list of Device Properties with a Device Class Name.
 device-class-name = 1*VCHAR
 ```
 
-* The Device Class MAY have one or more Device instance
-* The Device Class MAY have one or several Property, called Class Property
+* The Device Class MAY have one or more Device instance.
+* The Device Class MAY have one or several Property, called Class Property.
 
-* The Device SHALL expose (provide access to) all Attributes, Pipes and Commands defined by its Device Class.
+* The Device SHALL have all Attributes, Pipes and Commands defined by its Device Class.
```
hardion: The meaning of sentence avoid to use Dynamic Attribute.

2/README.md
```diff
@@ -142,38 +140,88 @@ The Device Interface is a list of Attributes, Pipes and Commands provided by a D
...
 ### Device lifecycle
 
 The Device lifecycle is:
-* Device Initialisation phase, when Device object is created and made visible in a Tango Controls system (see [RFC-1](), [RFC-8]())
-* Device Operation phase, when Device Interface MUST be available to actors of the Tango Controls system.
-* Device Destruction phase, when Device SHOULD be unregistered from the Tango Control system and Device Interface MAY not
-  be available within the Tango Controls system. 
+* Device MUST be in one of two phases: Exported or Not Exported.
+* In the Exported phase, Device Interface MUST be accessible to actors of the Tango Controls System.
+* In the Not Exported phase, Device Interface MUST NOT be accessible to actors of the Tango Controls System.
```
hardion: Device Interface is not defined. Simply use "a Device MUST ...

2/README.md
hardion: OK we can have it both but I would rephrase it:
```
* The Device Server MUST manage the Device lifecycle.
```

2/README.md
```diff
@@ -142,38 +140,88 @@ The Device Interface is a list of Attributes, Pipes and Commands provided by a D
...
 
 The device lifecycle MUST be implemented by a Device Server.
 
-### Device State
+### Init command
+
+The Init command purpose is to re-initialise the Device.
+
+* The Init command, when called, SHALL in a sequence:
+  * gracefully un-initialise the Device (free any dynamic resources allocated to the Device)
```
hardion:
```
... (free any dynamic resources allocated to the Device)
```
is implementation dependant.

You can reinforce the spec by saying "returning to a state before its first initialisation" (state as in english semantic not State the attribute).

2/README.md
```diff
@@ -142,38 +140,88 @@ The Device Interface is a list of Attributes, Pipes and Commands provided by a D
...
+  * initialise the Device
```
hardion: I would reinforce the role of Property in Init
```
... using the values of its Properties
```

2/README.md
```diff
@@ -142,38 +140,88 @@ The Device Interface is a list of Attributes, Pipes and Commands provided by a D
...
+
+### Device State and Status
+
+* The *data type* of State Attribute MUST be `DevState`. The *read value* of Status attribute MUST be 
+  of `DevState` data type.
+* The State attribute's *writable* metadata MUST be `READ`.
+* The *data format* of State Attribute MUST be `SCALAR`.
```
hardion, line 170: For the command all of these spec are in the Reserved Command

2/README.md
```diff
@@ -142,38 +140,88 @@ The Device Interface is a list of Attributes, Pipes and Commands provided by a D
...
+* The *read value* of State Attribute SHOULD reflect the context of a running Device. If Device is related to hardware
+  or a virtual entity or some process, the value of State attribute SHOULD reflect the state of the entity or process.
```
hardion: It is a good statement. What are the "else" if not hardware, virtual entity nor process?

2/README.md
```diff
@@ -142,38 +140,88 @@ The Device Interface is a list of Attributes, Pipes and Commands provided by a D
...
+
+* The *data type* of Status attribute MUST `DevString`. The *read value* of Status attribute MUST be 
+  of `DevString` data type.
+* The Status attribute's `writable` metadata MUST be `READ`. 
+* The `data format` of the Status attribute MUST be `SCALAR`.
+
+* A *read value* of Status Attribute SHOULD provide status information related to the Device.
```
hardion: You simplify:
```
The Status Attribute SHOULD provide status information related to the Device. 
```

2/README.md
```diff
@@ -142,38 +140,88 @@ The Device Interface is a list of Attributes, Pipes and Commands provided by a D
...
+* The *read value* of State Attribute SHOULD reflect the context of a running Device. If Device is related to hardware
+  or a virtual entity or some process, the value of State attribute SHOULD reflect the state of the entity or process.
+
+* The *data type* of Status attribute MUST `DevString`. The *read value* of Status attribute MUST be 
+  of `DevString` data type.
+* The Status attribute's `writable` metadata MUST be `READ`. 
+* The `data format` of the Status attribute MUST be `SCALAR`.
+
+* A *read value* of Status Attribute SHOULD provide status information related to the Device. 
+* Default implementation SHOULD set Status Attribute's  *read value* to indicate the Device State.
```
hardion: Default implementation => By default the Status Attribute SHOULD indicate the Device State

Sorry I may be picky with the word implementation but it so easy to end up in the trap of telling how and not the what.

2/README.md
```diff
@@ -142,38 +140,88 @@ The Device Interface is a list of Attributes, Pipes and Commands provided by a D
...
+  or a virtual entity or some process, the value of State attribute SHOULD reflect the state of the entity or process.
+
+* The *data type* of Status attribute MUST `DevString`. The *read value* of Status attribute MUST be 
+  of `DevString` data type.
+* The Status attribute's `writable` metadata MUST be `READ`. 
+* The `data format` of the Status attribute MUST be `SCALAR`.
+
+* A *read value* of Status Attribute SHOULD provide status information related to the Device. 
+* Default implementation SHOULD set Status Attribute's  *read value* to indicate the Device State. 
+* The Status default behaviour MAY be re-implemented with a user-code in Device Server to provide
```
hardion: Redundant with line 177.
Also user-code is another word for implementation :-)

2/README.md
```diff
@@ -142,38 +140,88 @@ The Device Interface is a list of Attributes, Pipes and Commands provided by a D
...
+* The Status Command MUST return the same value as would be read from the Status Attribute at the time of the command execution.
 
-* Writing to a Device's Attribute or to a Device's Pipe  
-* Calling a Device's Command
+### State Machine
 
-Device MAY not execute (block) the attribute write or the pipe write operation or the command execution code for certain States.
-In such a case the the Device object SHALL respond with throwing an exception.  [TO-DO: specify which exception]
-The way how a Device object respond to the above calls define a Device State Machine.
+* If not implemented explicitly by the Device Server, State Attribute *read value* SHALL be set to `UNKNOWN`.
```
hardion: I won't sleep well tonight...

```
By default the State Attribute value SHALL be UNKNOWN.
```

2/README.md
```diff
@@ -142,38 +140,88 @@ The Device Interface is a list of Attributes, Pipes and Commands provided by a D
...
 
-* Writing to a Device's Attribute or to a Device's Pipe  
-* Calling a Device's Command
+### State Machine
 
-Device MAY not execute (block) the attribute write or the pipe write operation or the command execution code for certain States.
-In such a case the the Device object SHALL respond with throwing an exception.  [TO-DO: specify which exception]
-The way how a Device object respond to the above calls define a Device State Machine.
+* If not implemented explicitly by the Device Server, State Attribute *read value* SHALL be set to `UNKNOWN`. 
+  It is RECOMMENDED to override this behaviour by an adequate implementation of the Device Server.
```
hardion: Redundant line 177

2/README.md
```diff
@@ -142,38 +140,88 @@ The Device Interface is a list of Attributes, Pipes and Commands provided by a D
...
+
+* If at least one of Device Attribute has the *quality* set to `ALARM` the *read value* of Device State attribute 
+  MUST be `ALARM`.
```
hardion: MUST to SHOULD

This is not a strong constraint in Tango

2/README.md
```diff
@@ -142,38 +140,88 @@ The Device Interface is a list of Attributes, Pipes and Commands provided by a D
...
+
+Value of Device's State attribute MAY impact how the Device object responds to:
```
hardion: The Value of the State Attribute MAY impact how the Device ~~object~~ responds to:

2/README.md
```diff
@@ -142,38 +140,88 @@ The Device Interface is a list of Attributes, Pipes and Commands provided by a D
...
+
+* Reading and writing of its Attribute or Pipe  
+* Calling a Command
+
+The Device MAY not execute (block) the Attribute or Pipe read or write operation or the command execution code 
+for individual States. In such a case the Device object SHALL respond with throwing a `DevFailed` exception.
```
hardion: regarding the State Attribute value. ...

hardion: Device ~~object~~

2/README.md
```diff
@@ -142,38 +140,88 @@ The Device Interface is a list of Attributes, Pipes and Commands provided by a D
...
+The way how a Device object responds to the above calls define a Device State Machine.
```
hardion: Device ~~object~~

2/README.md
```diff
@@ -142,38 +140,88 @@ The Device Interface is a list of Attributes, Pipes and Commands provided by a D
...
+admin-device-name = domain "/" family "/" member
+domain = %i"dserver"
+family = 1*VCHAR
+member = 1*VCHAR
+``` 
 
+`admin-device-name` MUST not be used for Devices of other than `DServer` Device Class.  
 
-### Logging 
+### Reserved Device Class Names
```
hardion: Good catch!!!

2/README.md
```diff
@@ -142,38 +140,88 @@ The Device Interface is a list of Attributes, Pipes and Commands provided by a D
...
-### Device Black Box
+### Reserved Device Names
 
+However it is possible to use any Device Name as it is stated in [Naming convention section](#naming-convention),
+some groups of names are reserved. One is Admin Device Name (see [RFC 8/Server](/8)), defined as follows: 
 
-### Events
+``` ABNF
+admin-device-name = domain "/" family "/" member
+domain = %i"dserver"
```
hardion, line 210: What is the %i ?

2/README.md
```diff
@@ -142,38 +140,88 @@ The Device Interface is a list of Attributes, Pipes and Commands provided by a D
...
 
-### Logging 
+### Reserved Device Class Names
 
+Devices of some of Device Classes are used to provide standard Tango Controls System services. 
+These Device Classes' Names SHOULD not be used for other purposes. Below is a list of reserved Device Class Names:
 
-### Cache and polling mechanism
+* `%i"DataBaseds"`, see [RFC 6/Database](/6)
+* `%i"TangoAccessControl"`
```
hardion, line 223: Does it matter if the Database Device is from this class ? Same for TangoAccessControl?
For example does the libtango9 rely on the class name TangoAccesControl to find its device?

2/README.md
piogor: Capitals are already there :-), not seen here as the GitHub does not update the code displayed here.

2/README.md
piogor: @hardion, you have convinced me, I will update accordingly.

2/README.md
piogor: ok, done :-)

2/README.md
piogor: Ok, I will remove the note and write as you suggest.

2/README.md
```diff
@@ -3,80 +3,224 @@ domain: rfc.tango-controls.org
...
+* The State Command MUST return the same value as would be read from the State Attribute at the time 
+  of the command execution.
+* The Status Command MUST return the same value as would be read from the Status Attribute at the time of the command execution.
+
+### State Machine
+
+* If not implemented explicitly by the Device Server, State Attribute *read value* SHALL be set to `UNKNOWN`. 
+  It is RECOMMENDED to override this behaviour by an adequate implementation of the Device Server.
+
+* If at least one of Device Attribute has the *quality* set to `ALARM` the *read value* of Device State attribute
```
piogor: @hardion, I think It is worth to check the real behaviour of `ALARM` state and update the paragraph accordingly,  before merging. So, I will do this.

lorenzopivetta: @piogor , it used to work like that, at least in Tango 9.2.5

bourtemb: > If at least one of Device Attribute has the *quality* set to `ALARM` the *read value* of Device State attribute  MUST be `ALARM`.

This is the default behaviour, but only when the computed state would otherwise be ON.
This behaviour can eventually be changed by user code too if the device server does not invoke Tango::DeviceImpl::dev_state(); at the end of its dev_state() method (code which is normally in a POGO protected area in C++).

piogor: Thanks, @bourtemb,  So, I propose to change it to this:
* If at given time State attribute *read value*  would be otherwise `ON` and at least one of Device Attribute has the *quality* set to `ALARM` the State *read value* SHALL be, by default, set to `ALARM`.  However, this behaviour MAY be overridden by user code in a Device Server implementation.

What do you think @bourtemb and @hardion?

lorenzopivetta: What about: "If at a given time the Device State is ON and at least one of Device Attribute has the quality set to ALARM, the Device State SHALL be set to ALARM. However..."

piogor: Thanks, @lorenzopivetta. I will do the following to rise fromality a bit:
"If at a given time the Device State is ON and at least one of Device Attribute has the *quality* set to ALARM,
  the Device State *read value* SHALL be set to ALARM."

lorenzopivetta: Not sure about explicitly stating "*read value*"... it somehow suggests that device State may be written too...

piogor: It follows RFC 4/Attribute terminology. Anyway, there is: " The State attribute's *writable* metadata MUST be `READ`." to assure the State being read-only.

2/README.md
```diff
@@ -3,80 +3,227 @@ domain: rfc.tango-controls.org
...
+* The Device SHALL have all Attributes, Pipes and Commands defined by its Class.
+ 
+### Device Interface
+
+The Device Interface is a list of Attributes, Pipes and Commands provided by a Device.
+
+* The Device MAY expose Attributes and/or Commands not defined by its Class. These are called 
+  Dynamic Attributes and/or Dynamic Commands respectively, see [RFC-XX]().
+
+* The Dynamic Attributes and Dynamic Commands MAY be added to the interface during Device Initialisation phase and/or
```
piogor: @hardion, Here I found I left ambiguous `Initialisation` and `Operation` phases. I will patch it.

2/README.md
hardion: I like the sentence of Lorenzo. At some point you have to trade off the accuracy with the readability.
Note that it's a SHOULD as it is implementation dependant. And instead to use the word implementation, just use "by default".
```
* By default the Device State SHOULD be in ALARM if at a given time the Device State is ON and at least one of Device Attribute has the quality set to ALARM.
```

2/README.md
piogor: I will remove the comma.

2/README.md
piogor: Now, synchronised in other paragraphs, too.

2/README.md
piogor: @hardion, do you mean not to introduce Dynamic Attribute term?

2/README.md
piogor: Ok, I will remove Device Interface term

2/README.md
piogor: Ok, I will rephrase.

2/README.md
piogor: Ok, will be pushed, soon.

2/README.md
piogor: done

2/README.md
piogor: done

2/README.md
piogor: done

2/README.md
piogor: Yes, I think it would be good to move of detail specification (argin/argout) of commands Init, State and Status  RFC  3/Command to here.

2/README.md
piogor: There could be some biological or even metaphysical entity ;-).
Do you think I should somehow change the sentence? I would improve it a bit to:  
"If Device is related to hardware
  or a virtual entity or some process, the value of State attribute SHOULD reflect the state of the hardware, entity or process."

2/README.md
piogor: I agree with you :-)

piogor: It is better to keep it clear we do not suggest way of implementation.

2/README.md
piogor: I will remove the sentence

2/README.md
lorenzopivetta: What about: "If a Device represents an equipment (hardware) or rather a software component, the value of Device State SHOULD reflect the state of the equipment, or software component". I'd avoid *entity* or *process* here...

2/README.md
hardion: "The Device have all attributes defined by its Class" means that there is no other way to define Attribute, Command and Pipe only by the Class. Which is wrong since we can define Dynamic Attribute.
But I have hard time to rephrase it.

2/README.md
piogor: Sorry for this. I will simplify.

2/README.md
piogor: I will remove.

piogor: Removed

2/README.md
piogor: done wiht the recent push.

2/README.md
piogor: done with the recent push

2/README.md
piogor: ABNF case insensitive: https://tools.ietf.org/html/rfc7405#section-2.1
In fact, it may be redundant here but I think we do not explicitly mention that names are case-insensitive.

2/README.md
piogor: You may be right, and it is more conventionality that these classes have their names. I will search through source code. However, if we define that Class is an association of its Attributes, Commands etc. with a Class Name we should reserve names related to standard system-wide services.

2/README.md
piogor: Maybe:
"The Device SHALL have at least al..."?

2/README.md
piogor: done

2/README.md
piogor: done

2/README.md
piogor: `TanogAccessControl` appears in tests,
A class name is checked for not being `DataBase` when locking is invoked. So, I suggest leaving these names here. I will just patch `DataBaseds` to be `DataBase`.

2/README.md
piogor: the paragraph is rephrased.

2/README.md
piogor: done

2/README.md
piogor: done

2/README.md
piogor: done

piogor:
Thanks, @hardion, @lorenzopivetta, @bourtemb, @gwen-soleil, @david-erb-maxiv  and all others (@mliszcz ), for all the comments and discussions. I think, with the most recent commits all are somehow addressed and ready for a (probably final) review.

PR #38, opened 2019-07-30
WIP Publisher-Subscriber protocol (RFC-12)

HelgaMond:
Tango Controls V9 specification for Publisher-Subscriber protocol (RFC-12).


12/README.md
```diff
@@ -0,0 +1,165 @@
...
+*Tango Event Callback* (callback) - a piece of code that *client library* executes when an update is received.
+
+## Definitions
+
+This section offers a number of ABNF formal definitions for entities repeatedly used below in this document. 
+
+```ABNF
+EVENT_TYPE = "QUALITY_EVENT" | "INTERFACE_CHANGE" | "PIPE" | "ATTR_CONF_EVENT" |  "CHANGE_EVENT" | "PERIODIC_EVENT" | "ARCHIVE_EVENT" | "USER_EVENT" | "HEARTBEAT"
+
+HEARTBEAT_ENDPOINT = "(upstream server FQDN):port"
```
Ingvord: These definitions should go into implementation as they are a part of ZMQ specific part

12/README.md
```diff
@@ -0,0 +1,165 @@
...
+*Tango Event* (event) - an update sent by *server*.
+
+*Tango Event Callback* (callback) - a piece of code that *client library* executes when an update is received.
+
+## Definitions
+
+This section offers a number of ABNF formal definitions for entities repeatedly used below in this document. 
+
+```ABNF
+EVENT_TYPE = "QUALITY_EVENT" | "INTERFACE_CHANGE" | "PIPE" | "ATTR_CONF_EVENT" |  "CHANGE_EVENT" | "PERIODIC_EVENT" | "ARCHIVE_EVENT" | "USER_EVENT" | "HEARTBEAT"
```
Ingvord: Here we should probably describe different events in more details e.g. 
```
EVENT = ATTRIBUTE_VALUE_CHANGE | ...
ATTRIBUTE_VALUE_CHANGE = such an event when Attribute's value (link to RFC attribute) changes as per configured in Device properties (link to properties) 
```

bourtemb: I think we should use "/" instead of "|" in ABNF to describe alternatives (https://tools.ietf.org/html/rfc5234#section-3.2)

12/README.md
```diff
@@ -0,0 +1,165 @@
...
+
+HEARTBEAT_ENDPOINT = "(upstream server FQDN):port"  
+
+EVENT_ENDPOINT = "(upstream server FQDN):port"
+
+HEARTBEAT_CHANNEL = "(upstream admin device FQDN)/heartbeat" 
+
+EVENT_CHANNEL = "(upstream admin device FQDN)/(attr).EVENT_TYPE[_idl5]"
```
bourtemb: [idl5_] is before the EVENT type I think and it's not the upstream admin device in this case, but the upstream device:
EVENT_CHANNEL = "(upstream device FQDN)/(attr).[idl5_]EVENT_TYPE"

12/README.md
```diff
@@ -0,0 +1,165 @@
...
+EventSubscriptionChange_ARGOUT-type = DevLong
+EventSubscriptionChange_ARGOUT-value = "(server library release ver)"
+```
+
+**NOTE**: Tango Server Library v9.3.3 force client to explicitly call ZmqEventSubscriptionChange to use ZMQ based implementation [RFC-13]: 
+ 
+```ABNF
+ZmqEventSubscriptionChange = Tango Command
+ZmqEventSubscriptionChangeArgument-type  = DevVarStringArr 
+ZmqEventSubscriptionChangeArgument-value = "[ upstream device, attribute, action ("subscribe|unsubscribe"), EVENT_TYPE]"
```
bourtemb: Passing "info" as parameter is a nice (kind of hidden) feature too.
ZmqEventSubscriptionChangeArgument-value = "[ upstream device, attribute, action ("subscribe" / "unsubscribe"), [idl5_]EVENT_TYPE]" / "info"

12/README.md
```diff
@@ -0,0 +1,165 @@
...
+
+- server heartbeat
+- client subscription confirmation
+- events client/server buffer
+
+* Client SHOULD try to re-subscribe once HEARTBEAT event is missing
+
+* Client MAY try to re-subscribe indefinitely
+
+* Server MAY cancel client's subscription if it does not confirm its interest i.e. EventConfirmSubscription was not executed
```
bourtemb: I would propose something like that.
Server MAY cancel client's subscription if it didn't confirm its interest in the last 600 seconds i.e. EventConfirmSubscription command was not sent by this client in the last 600 seconds.

12/README.md
```diff
@@ -0,0 +1,165 @@
...
+
+* Client MAY use provided heartbeat/event channels to identify event source 
+
+### Subscription 
+
+* Server MUST send events via corresponding EVENT_CHANNEL
+
+* Server MUST send HEARTBEAT event every 9 seconds
+
+* Client MUST execute EventConfirmSubscription command every 600 seconds to indicate its interest in subscription
```
bourtemb: In practice, the C++ and Java client libraries are sending this command every 200 seconds because what is important is for the upstream server to receive it **before** 600 seconds.

12/README.md
```diff
@@ -0,0 +1,165 @@
...
+
+* Client MAY cancel its subscription by executing EventSubscriptionChange on upstream admin 
+
+```abnf
+
+EventSubscriptionChange_ARGIN-value = "[ (upstream device), (attribute), "unsubscribe", EVENT_TYPE]"
+
+```
+
+* Server MUST stop to send events to client whose subscription is canceled
```
bourtemb: Sorry if I bring some confusion. 
In the current implementation . When using ZMQ (at least) to send events, server MAY continue to send events to the client whose subscription is canceled if there is another client, interested in the same event type (same event type with same level of compatibility) which sent an event confirm subscription command in the last 600 seconds and if the client whose subscription is canceled didn't close the ZMQ event channel connection yet.

12/README.md
```diff
@@ -0,0 +1,165 @@
...
+
+* Client SHOULD try to re-subscribe once HEARTBEAT event is missing
+
+* Client MAY try to re-subscribe indefinitely
+
+* Server MAY cancel client's subscription if it does not confirm its interest i.e. EventConfirmSubscription was not executed
+
+* Client/Server SHOULD implement (in/out)coming events buffer to prevent overload 
+
+* Client/Server MUST prevent crash in case events overflow
```
bourtemb: Clients SHOULD be notified in case of events overflow (e.g.: In the current C++ implementation at least, the client callback is executed with an API_MissedEvents error event when the client library detected that some events got lost. There is a mechanism with some counters associated with each event to help detecting this kind of problems).

12/README.md
```diff
@@ -0,0 +1,165 @@
...
+```
+
+## Runtime requirements
+
+Client and server are up and running. Server is reachable from client i.e. may communicate using Request-Reply protocol [RFC-10]. 
+
+## Publisher-Subscriber protocol
+
+![Subscription_1.png](images/Subscription_1.png)
+
```
bourtemb, line 107: I don't understand the failed part of the diagram on the left.

12/README.md
```diff
@@ -0,0 +1,165 @@
...
+* Server library MUST provide an API to publish events of EVENT_TYPE
+
+* Client SHOULD process events
+
+* Client library SHOULD provide user an API to register user defined event callbacks  
+
+### Negotiation
+
+![Subscription_2.png](images/Subscription_2.png)
```
bourtemb, line 120: I would add some words over every arrow to avoid confusion.

bourtemb (commented):
Nice first draft.
Please have a look at my comments which I hope will be helpful and won't bring too much confusion.

12/README.md
bourtemb: Please also note that Java clients do not currently use EventConfirmSubscription command to confirm their subscriptions but they are using ZmqEventSubscriptionChange command regularly instead for the same purpose.

Ingvord:
> Nice first draft.
> Please have a look at my comments which I hope will be helpful and won't bring too much confusion.

@bourtemb , thanks a lot for the comments! These will definitely help!

12/README.md
```diff
@@ -0,0 +1,165 @@
...
+x
+This Specification is a [free and open standard](http://www.digistan.org/open-standard:definition) and is governed by the Digital Standards
+Organization's [Consensus-Oriented Specification System](http://www.digistan.org/spec:1/COSS).
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119).
+
+## Introduction
+
+This document describes the Tango Publisher-Subscriber protocol. It is a relationship between Tango client (client) and Tango server (server). Based on criteria negotiated as part of a subscription, Tango client will receive updates. Beyond a set of basic requirements, various refinements are addressed. These refinements include: periodicity of updates, filtering out updates based on Tango server configuration and fault tolerance.
```
hardion: You can move this part before the preamble, without the introduction title (see Command for example).
It's also the right place to make the reference to the other RFC i.e:
```
See also: 2/Device X/Request-Reply
```

12/README.md
Ingvord: We have added text to the arrow

12/README.md
Ingvord: Done

Ingvord:
Add message parsing info i.e. CORBA serializing/deserializing

Ingvord:
~Add EVENT_INFO~Done in 6e306ea

12/README.md
```diff
@@ -0,0 +1,165 @@
...
+This Specification is a [free and open standard](http://www.digistan.org/open-standard:definition) and is governed by the Digital Standards
+Organization's [Consensus-Oriented Specification System](http://www.digistan.org/spec:1/COSS).
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119).
+
+## Introduction
+
+This document describes the Tango Publisher-Subscriber protocol. It is a relationship between Tango client (client) and Tango server (server). Based on criteria negotiated as part of a subscription, Tango client will receive updates. Beyond a set of basic requirements, various refinements are addressed. These refinements include: periodicity of updates, filtering out updates based on Tango server configuration and fault tolerance. 
+
+## Basic Concepts
```
hardion, line 26: This paragraph need some attention as many keyword should be already defined in other RFC.

12/README.md
hardion: @bourtemb is right

12/README.md
hardion: @HelgaMond @Ingvord Can you check the comment of @bourtemb

12/README.md
```diff
@@ -0,0 +1,165 @@
...
+EVENT_ENDPOINT = "(upstream server FQDN):port"
+
+HEARTBEAT_CHANNEL = "(upstream admin device FQDN)/heartbeat" 
+
+EVENT_CHANNEL = "(upstream admin device FQDN)/(attr).EVENT_TYPE[_idl5]" 
+```
+
+## Runtime requirements
+
+Client and server are up and running. Server is reachable from client i.e. may communicate using Request-Reply protocol [RFC-10].
```
hardion, line 102: Client and Server...
and they should be defined in RFC-10. (I open an issue)
#82

12/README.md
hardion: I open an issue to define the FQDN in their respective RFCs #80

12/README.md
```diff
@@ -0,0 +1,165 @@
...
+
+## Runtime requirements
+
+Client and server are up and running. Server is reachable from client i.e. may communicate using Request-Reply protocol [RFC-10]. 
+
+## Publisher-Subscriber protocol
+
+![Subscription_1.png](images/Subscription_1.png)
+
+The main goal of the this protocol implementation is to allow client receive events from server:
```
hardion: The main goal of the this protocol implementation is to allow a Client to receive events from a Server:

12/README.md
```diff
@@ -0,0 +1,165 @@
...
+
+```ABNF
+EventSubscriptionChange = Tango Command
+EventSubscriptionChange_ARGIN-type  = DevVarStringArr 
+EventSubscriptionChange_ARGIN-value = "[ (upstream device), (attribute), "subscribe", EVENT_TYPE]"
+EventSubscriptionChange_ARGOUT-type = DevLong
+EventSubscriptionChange_ARGOUT-value = "(server library release ver)"
+```
+
+**NOTE**: Tango Server Library v9.3.3 force client to explicitly call ZmqEventSubscriptionChange to use ZMQ based implementation [RFC-13]:
```
hardion: The note and the ABNF below should go to  #39 
but here should be specify that a specific protocol should use their own set of command on the Admin Device (unfortunately):

```ABNF
XyzEventSubscriptionChange = Tango Command
````
Where Xyz is the code name of the specific P/S protocol.

Comment (hardion) [reply 12/README.md]: Another question: Are to one to one compatible? ZmqEventSubscriptionChangeArgument-value compatible with EventSubscriptionChange_ARGIN-value

Comment (hardion) [reply 12/README.md]: Another question: Is the command EventSubscriptionChange only for notifyd?

File: 12/README.md (hunk 4 above)
Comment (hardion) [reply 12/README.md]: @bourtemb should we define "info" in this case?

File: 12/README.md [hunk 16]
```diff
@@ -0,0 +1,165 @@
...
+ZmqEventSubscriptionChange = Tango Command
+ZmqEventSubscriptionChangeArgument-type  = DevVarStringArr 
+ZmqEventSubscriptionChangeArgument-value = "[ upstream device, attribute, action ("subscribe|unsubscribe"), EVENT_TYPE]"
+ZmqEventSubscriptionChangeOutput-type = DevVarLongStringArr
+ZmqEventSubscriptionChangeOutput-value = "[(server library release ver), (upstream device idl ver), (zmq_sub_event_hwm), (rate), (ivl), (zmq release ver); (heartbeat_endpoint), (event_endpoint), n*(alternate_heartbeat_endpoint), n*(alternate_event_endpoint),(heartbeat_channel), (event_channel)]"
+```
+
+* Client MUST execute admin's EventSubscriptionChange command to establish Publisher-Subscriber relationship
+
+* Client SHOULD establish tcp? connection to heartbeat/event endpoints
```
hardion: Client SHOULD establish the connection to the  heartbeat/event endpoints with the protocol requested by the Server

12/README.md
hardion: @bourtemb So it is RECOMMENDED to send this confirmation every 200 s

12/README.md
hardion: @bourtemb This is interesting. Although I am not to follow the purpose. It is due to how Zmq works or the definition of what is a Client (computer / program / thread / socket ...)

12/README.md
hardion: Isn't already specify in line 131 in Subscription? Maybe every thing about the heartbeat and confirmation should in in this paragraph.

12/README.md
hardion: I prefer what @bourtemb describes than specifying to avoid crash ;-)

12/README.md
```diff
@@ -0,0 +1,165 @@
...
+- client subscription confirmation
+- events client/server buffer
+
+* Client SHOULD try to re-subscribe once HEARTBEAT event is missing
+
+* Client MAY try to re-subscribe indefinitely
+
+* Server MAY cancel client's subscription if it does not confirm its interest i.e. EventConfirmSubscription was not executed
+
+* Client/Server SHOULD implement (in/out)coming events buffer to prevent overload
```
hardion: * Client/Server SHOULD buffer the (in/out)coming events  to prevent overload  and overflow

hardion (changes requested):
The specification sounds almost complete and now can refer to other complete RFCs. 
Maybe the actions in the protocol should enumerated i.e First the Client call Event..., Secondly the Server answers ...

Do we have the source of the png image?

12/README.md
bourtemb: > @bourtemb should we define "info" in this case?

I think it would be useful. This feature is really useful to diagnose event issues.

12/README.md
bourtemb: This is only due to the way the unsubscription is currently implemented. When a Tango client unsubscribes for a given event, there might be a short delay before the ZMQ unsubscription is actually sent on the network, so the server MIGHT continue to send some events until it really received the ZMQ unsubscription command.
The events received by the client between the unsubscribe_event() Tango call and the ZMQ event unsubscription reception on the server-side will simply be ignored by the Tango client.

hardion: OK you describe also "another client"

bourtemb: Right you are! What I wrote before is indeed confusing... But I warned at the beginning of the first comment: ;-)
> Sorry if I bring some confusion.

No need for "Another client"!

HelgaMond:
@hardion Thank you very much for your comments. But it seems that most of them were done to the outdated version...  Could you, please, check your version and update your comments? Thank you!

HelgaMond:
Dear @hardion . I have made minor changes in the RFC. Please, update before the review. Thank you!

12/README.md
```diff
@@ -0,0 +1,189 @@
...
+
+EVENT_ERROR = reason severity desc origin
+
+EVENT_TYPE = "QUALITY_EVENT"/"INTERFACE_CHANGE"/"PIPE"/"ATTR_CONF_EVENT"/"CHANGE_EVENT"/"PERIODIC_EVENT"/"ARCHIVE_EVENT"/"USER_EVENT"/"HEARTBEAT"
+
+EVENT_CHANNEL = channel
+
+HEARTBEAT_CHANNEL = channel
+
+channel = string ; impementation specific
```
bourtemb:
```suggestion
channel = string ; implementation specific
```

12/README.md
```diff
@@ -0,0 +1,189 @@
...
+
+## Runtime requirements
+
+Client and server are up and running. Server is reachable from client i.e. may communicate using Request-Reply protocol [RFC-10]. 
+
+## Publisher-Subscriber protocol
+
+![Subscription_1.png](images/Subscription_1.png)
+
+The main goal of the this protocol implementation is to allow a Client receive events from a Server:
```
bourtemb:
```suggestion
The main goal of this protocol implementation is to allow a Client to receive events from a Server:
```

12/README.md
```diff
@@ -0,0 +1,189 @@
...
+
+DATA_READY_DATA = attribute_name data_type int:counter
+
+INTERFACE_CHANGE_DATA = *ATTRIBUTE_PROPERTIES *COMMAND_META_INFO bool:deviceStarted
+
+ATTRIBUTE_DATA = ATTRIBUTE_DEFINITION ; Attribute definition from RFC-4 
+
+EVENT_ERROR = reason severity desc origin
+
+EVENT_TYPE = "QUALITY_EVENT"/"INTERFACE_CHANGE"/"PIPE"/"ATTR_CONF_EVENT"/"CHANGE_EVENT"/"PERIODIC_EVENT"/"ARCHIVE_EVENT"/"USER_EVENT"/"HEARTBEAT"
```
bourtemb: QUALITY_EVENT is commented as deprecated in cppTango:
https://github.com/tango-controls/cppTango/blob/40aecac1388b5190ec281d99659d170f35a2d5a2/cppapi/server/tango_const.h.in#L1106

```suggestion
EVENT_TYPE = "INTERFACE_CHANGE"/"PIPE_EVENT"/"ATTR_CONF_EVENT"/"CHANGE_EVENT"/"PERIODIC_EVENT"/"ARCHIVE_EVENT"/"USER_EVENT"/"HEARTBEAT"
```

bourtemb: I don't think the Tango client API lets you subscribe to HEARTBEAT events directly.
This HEARTBEAT event subscription is done implicitly when subscribing to another event (it is a low level subscription).
HEARTBEAT as EVENT_TYPE is never used in the SUBSCRIBER_INFO which is listed above.
This is why HEARBEAT is not listed in cppTango enumType enumeration:
https://github.com/tango-controls/cppTango/blob/40aecac1388b5190ec281d99659d170f35a2d5a2/cppapi/server/tango_const.h.in#L1104-L1115

12/README.md
```diff
@@ -0,0 +1,189 @@
...
+ It is a relationship between Tango client (Client) and Tango server (Server). Based on criteria negotiated as part of a subscription, Tango client will receive updates. Beyond a set of basic requirements, various refinements are addressed. These refinements include: periodicity of updates, filtering out updates based on Tango server configuration and fault tolerance. 
+
+
+## Basic Concepts
+
+This documents refers to other Tango Controls RFCs, corresponding link is always given in this case. Here is given a brief description of the main concepts. This document targets arbitrary software developer(s) who are to implement *Tango Client Library* and *Tango Server Library*.
+
+*Tango Server* (Server) - a process that runs on a machine.
+
+*Tango Device* (Device) - a thread within a *Tango Server* that implements *Tango Class* i.e. has *Tango Commands, Attributes, Pipes, Properties*  etc.
```
bourtemb: A Tango Device is a thread?
I guess we should refer to other RFCs for most of these definitions, as @hardion already commented.

12/README.md
```diff
@@ -0,0 +1,189 @@
...
+
+*Tango Admin Device* (Admin) - a pre-defined *Tango Device* i.e. a *Tango Device* with fixed number *Tango Commands, Attributes, Pipes* [RFC-3, RFC-4,RFC-?]  etc.
```
bourtemb: a Tango Device with fixed number of Tango commands, Attributes, Pipes, ...?
Most of the Tango devices have a fixed number of Tango commands, attributes, pipes, etc... but they are not necessarily Tango Admin Devices.
We should refer to RFC-8 (or RFC-2 you're right but RFC-8 seems to be dedicated to the admin device) for the definition of the admin device.

12/README.md
```diff
@@ -0,0 +1,189 @@
...
+
+For more details see [RFC-2](https://github.com/tango-controls/rfc/blob/draft/2/README.md)
```
bourtemb:
```suggestion
For more details see [RFC-2](https://github.com/tango-controls/rfc/blob/draft/2/Device.md)
```

HelgaMond:
@bourtemb Thank you very much for the review. I have made the changes. If RFC is in "Raw" status I have left it without links so that not to have 404 pages, e.g. links to RFC-10, RFC-8 or RFC-13.

PR #39, opened 2019-07-30
WIP Publisher-Subscriber protocol - ZeroMQ implementation (RFC-13)

HelgaMond:
Tango Controls V9 specification: Publisher-Subscriber protocol - ZeroMQ implementation (RFC-13).


hardion:
@HelgaMond @Ingvord 
Thanks for the last commit.
I did a quick review and I am afraid you are going too deep into the libtango API. It sounds more like the libtango documentation. You should imagine someone rewriting the ZeroMQ implementation in a total different way but still be compatible with Tango.
Maybe doing RFC 12 before would force to abstract the event system. And this case you can say which protocol of ZeroMQ Tango use, complete a sequence diagram, describe the new command the dserver should have, what is the specific zeromq frame structure on the wire (if any and needed). 
One another tip, you can use subscriber instead of client in the text.

I hope to help and not demotivate you. This is not an easy job!

Ingvord:
@hardion , 

Thank you for your fast response. 

It is just the beginning. Basically it includes only client part. So diagrams etc will come later.

>  You should imagine someone rewriting the ZeroMQ implementation in a total different way but still be compatible with Tango.

This is exactly how we do it. Unfortunately I can not  image anyone now who will rewrite anything in Tango without rewriting it from scratch. So our main priority is to describe current implementation as detailed as required to understand what is going on and how that can be re-implemented.

> Maybe doing RFC 12 before would force to abstract the event system. 

RFC-12 has already quite abstract description. Here we took abstractions from the existing code base indeed. I think that this specification must describe what is currently happening within Tango when we talk about Tango Event System. That is not possible to describe it without touching actual C++/Java implementations as we are talking about the _implementation_.

What we have done so far: we have analyzed C++/Java code, compare it, highlight the main points.  What you see in this RFC is not functions from the code, but more or less abstract entities used in the implementation. As for us these abstractions are essential for current Tango PUB-SUB protocol. Based on this we (Tango Community) may define the requirements, and propose a new RFC for future implementations. That RFC will be non-implementation agnostic.

If we abstract RFC-13 too much, we are afraid, we can lose some essential details. Because current code base is loosely tighten with technologies being used i.e. ZMQ or NotifD.

Ingvord:
~Specify that message is serialized/deserialized using CORBA  streams~ Done in 5324e1a

PR #43, opened 2019-08-09
Updating the editors table with current Editors - 1…

piogor:
…90808

This is an update after the meeting 08-08-2019

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ The process to add or change an RFC is the following:
 Short Name             | Title                                                         | Type     | Status     | Editor
 -----------------------|---------------------------------------------------------------|----------|------------|-------
 [RFC-1](1/README.md)   | The Tango control system                                      | Standard | Draft      | Lorenzo Pivetta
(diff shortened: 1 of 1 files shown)


hardion (commented):
The difference between Raw and Draft is a bit ambiguous in our case since the implementation of Tango9 came before the specification. So in theory we shave already a mockup of Tango9 and now we should write the draft of the specification.

I would stick to DRAFT but please ask around what the people think.


See exact definition:

Draft Specifications: When raw specifications can be demonstrated, they become draft specifications. Changes to draft specifications should be done in consultation with users. Draft specifications are contracts between the editors and implementers.


Raw Specifications: All new specifications are raw specifications. Changes to raw specifications can be unilateral and arbitrary. Those seeking to implement a raw specification should ask for it to be made a draft specification. Raw specifications have no contractual weight.

hardion:
I can merge if you split the editors change from the Draft to Raw changes.

piogor:
@hardion, so I made this PR with editors update only :). I will create another one with a proposition of changing of the statuses.

PR #44, opened 2019-08-09
Specification statuses update 

piogor:
This is a proposition of update to the statuses. 
It marks the RFC currently in progress or not yet started as raw and left Draft for RFC-3 only.
However, please look on this comment: https://github.com/tango-controls/rfc/pull/43#pullrequestreview-273016562 by @hardion.

@hardion, the PR contains the commit from #43, to avoid conflicts if we decide to merge it after #43.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,23 +11,23 @@ The process to add or change an RFC is the following:
 
 Short Name             | Title                                                         | Type     | Status     | Editor
 -----------------------|---------------------------------------------------------------|----------|------------|-------
-[RFC-1](1/README.md)   | The Tango control system                                      | Standard | raw        | Lorenzo Pivetta
-[RFC-2](2/README.md)   | The device object model                                       | Standard | raw        | Vincent Hardion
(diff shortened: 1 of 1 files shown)


bourtemb (commented):
Fine with me to put only RFC-3 as Draft and the others as Raw for the moment.
Just to be picky, I would put "Raw" instead of "raw" for consistency. ;-)

piogor:
In replly to: https://github.com/tango-controls/rfc/pull/43#pullrequestreview-273016562 ,

I think, currently, only the Command specification can be (and is) demonstrated by Tango9.
Other specifications, not yet accepted to be merged, probably are missing something that is demonstrated by the Tango9 (this is the reason for not being yet merged). 
For example, Tango9 implements some specification of Device model, however, at this point, I may be wrong in what I have written and in this sense RFC-2 is not demonstrated by Tango9.

hardion:
I can understand the arguments. 
Actually the MR is for the transition from Raw to Draft.

PR #45, opened 2019-08-26
WIP Database (RFC-6)

david-erb:
add stub README.md for RFC-6

### DIFF
--- a/6/README.md
+++ b/6/README.md
@@ -51,7 +51,7 @@ Runtime implementations of the Database in conforming Tango implementations MUST
 * define persistent configuration properties to be used by a camera every time it starts up
 * allow a subscriber or client to get connect to a stepper motor
 * define properties for all instances of a class of vaccuum pump
-* check if a electrometer is online or not
+* provide the network ID of a device such as an electrometer, so a user can check if it is online or not 
 * a user would like to rollback the configuration of a Device when its initialization is failing
 * remove "empty" Devices from Database to clean it up
 * get all Devices of a given class of temperature gauges
@@ -62,7 +62,15 @@ Runtime implementations of the Database in conforming Tango implementations MUST
 ## Specification
 The database SHALL provide a means of defining and querying devices.
 
+As the Tango database service is a central service of the Tango control system, 
+it SHALL be designed to be highly available and support a high level of client requests 
+(e.g we have peaks of 3500 SQL requests/ s at SOLEIL on the accelerators control system).
+
 The database service SHALL be implemented as a Tango Device Server.
+
+The database service SHALL be implemented with:
+- a frontend: as Tango Device Server
+- a backend: a query and persistency engine (i.e. SQL or NoSQL engine), or a lighter ( i.e. reduced set of functionalities) engine with a single file per device server.
 
 The Database Device SHALL have no Class Properties or Device Properties.
 
@@ -1021,28 +1029,28 @@ The following set of limitations exist:
 * in case of several Device servers running on the same host, the user must manually manage a list of already used network ports
 
 ### File syntax
-The file is an ASCII file and SHALL follow a well-defined syntax with predefined keywords.
+The file is an ASCII file.  It syntax is defined completely in .
 
 Generally, comment starts with the ’#’ character and a blank line is ignored.
 
-#### Devices definition.
+#### File syntax: Devices definition.
 DEVICE is the keyword to declare a device(s) definition sequence. The general syntax SHALL be   
 ```<DS name>/<inst name>/DEVICE/<Class name>: dev1,dev2,dev3  ```
 Device(s) name can follow on next line if the last line character is ’\’
 
-#### Device property definition. 
+#### File syntax: Device property definition. 
 The general device property SHALL be  
 ```<device name>-><property name>: <property value>```  
 In case of array, the array element delimiter is the character ’,’. Array definition can be splitted on several lines if the last line character is ’\’. Allowed characters after the ’:’ delimiter are space, tabulation or nothing.
 
 A device string property with special characters (spaces). The ’’ character is used to delimit the string
 
-#### Device attribute property definition. 
+#### File syntax: Device attribute property definition. 
 The general device attribute property syntax SHALL be   
 ```<device name>/<attribute name>-><property name>: <property value>```  
 Allowed characters after the ’:’ delimiter are space, tabulation or nothing.
 
-#### Class property definition. 
+#### File syntax: Class property definition. 
 The general class property syntax SHALL be  
 ```CLASS/<class name>-><property name>: <property value>```  
 
@@ -1051,9 +1059,9 @@ The ’’ characters around the property value are mandatory due to the ’/’
 
 
 ## Related documentation
-[Tango Controls Documentation 9.3.3](https://buildmedia.readthedocs.org/media/pdf/tango-controls/latest/tango-controls.pdf)
+[Tango Controls Documentation](https://buildmedia.readthedocs.org/media/pdf/tango-controls/latest/tango-controls.pdf)
 
-[PyTango Documentation Release 9.3.1](https://readthedocs.org/projects/pytango/downloads/pdf/stable/pytango.pdf)
+[PyTango Documentation Release](https://readthedocs.org/projects/pytango/downloads/pdf/stable/pytango.pdf)
 
 [DataBase Tango Cpp Class](http://www.esrf.eu/computing/cs/tango/tango_doc/ds_doc/tango-ds/System/dbase/FullDocument.html)
 
--- a/6/README.md
+++ b/6/README.md
@@ -1,0 +1,47 @@
+---
+domain: rfc.tango-controls.org
+shortname: 6/Database
+name:Database
+status: draft
+editor: 
+contributors: [David Erb](mailto:david.erb@maxiv.lu.se)
+---
+
+This document describes the Tango Database specification.
+
+
+## Preamble
+
+Copyright (c) 2019 Tango Controls
+
+This Specification is free software; you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation; either version 3 of the License, or (at your option) any later version. This Specification is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details. You should have received a copy of the GNU General Public License along with this program; if not, see <http://www.gnu.org/licenses>.
+
+This Specification is a [free and open standard](http://www.digistan.org/open-standard:definition) and is governed by the Digital Standards Organization's [Consensus-Oriented Specification System](http://www.digistan.org/spec:1/COSS).
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119).
+
+
+## Tango Database specification
+
+This specification is intended to formally document the Tango Database. 
+Runtime implementations of the Database in conforming Tango implementations
+MUST follow this specification.
+
+
+### Goals
+
+TBD
+
+### Use Cases
+
+Some example use cases of an Attribute are:
+* TBD
+
+## Specification
+
+The Database has:
+* TBD
+
+### Database definition
+
+TBD
--- a/6/README.md
+++ b/6/README.md
@@ -33,15 +33,15 @@ Runtime implementations of the Database in conforming Tango implementations MUST
 ### Goals of the Database
 * provide a means of defining and querying devices.
 * provide means for device servers to be configured from central control
-* serves as nameserver such that such that the Devices are accessible by name
+* serve as nameserver such that the Devices are accessible by name
 
 ### Features of the Database
 
-* stores configuration data (properties) used at startup of a device server 
-* acts as nameserver by storing the dynamic network addresses
-* acts as permanent store of dynamic settings which need to be memorized
-* ensures the uniqueness of device name and of aliases
-* keeps list of controlled servers which are to be started and run on a particular host
+* store configuration data (properties) used at startup of a device server 
+* act as nameserver by storing the dynamic network addresses
+* act as permanent store of dynamic settings which need to be memorized
+* ensure the uniqueness of device name and of aliases
+* keep list of controlled servers which are to be started and run on a particular host
 * limited set functionality provide by file as alternative to network database
 
 ### Use Cases of the Database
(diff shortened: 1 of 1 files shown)


6/README.md
```diff
@@ -0,0 +1,275 @@
...
+## Tango Database specification
+
+This specification is intended to formally document the Tango Database. 
+Runtime implementations of the Database in conforming Tango implementations MUST follow this specification.
+
+
+### Goals of the Database
+
+* provide means for device servers to be configured from central control
+* serves as nameserver such that clients are able to find a device server's network location at runtime by name
```
hardion: "client" is a double sword word which people understand the tango client library. So the sentence make sense but are limiting too. A "subscriber" may also want to find the device.

Why not "serve as name server such that the Device are accessible by name"

6/README.md
```diff
@@ -0,0 +1,275 @@
...
+
+### Use Cases of the Database
+
+* stores configuration data (properties) used at startup of a device server 
+* acts as nameserver by storing the dynamic network addresses
+* acts as permanent store of dynamic settings which need to be memorised
+* ensures the uniqueness of device name and of aliases
+* keeps list of controlled servers for Starter service
+* limited set functionality provide by file as alternative to network database
```
hardion, line 45: You can be more specific in your use cases i.e:
 stores the last user setting point of the current provided by a power supply.

6/README.md
```diff
@@ -0,0 +1,275 @@
...
+* provide means for device servers to be configured from central control
+* serves as nameserver such that clients are able to find a device server's network location at runtime by name
+
+### Use Cases of the Database
+
+* stores configuration data (properties) used at startup of a device server 
+* acts as nameserver by storing the dynamic network addresses
+* acts as permanent store of dynamic settings which need to be memorised
+* ensures the uniqueness of device name and of aliases
+* keeps list of controlled servers for Starter service
```
hardion: Avoid Starter word. This is an undefined term which would probably required an RFC (Not the priority at this point of time)

6/README.md
```diff
@@ -0,0 +1,275 @@
...
+
+* provide means for device servers to be configured from central control
+* serves as nameserver such that clients are able to find a device server's network location at runtime by name
+
+### Use Cases of the Database
+
+* stores configuration data (properties) used at startup of a device server 
+* acts as nameserver by storing the dynamic network addresses
+* acts as permanent store of dynamic settings which need to be memorised
+* ensures the uniqueness of device name and of aliases
```
hardion: This sounds more like a feature than a use case.

6/README.md
```diff
@@ -0,0 +1,275 @@
...
+* keeps list of controlled servers for Starter service
+* limited set functionality provide by file as alternative to network database
+
```
hardion, line 46: A user would like to rollback the configuration of a device when its initialisation is failing

6/README.md
```diff
@@ -0,0 +1,275 @@
...
+
+### General Specification
+The database service shall be implemented as a Tango Device Server.
+
+In the Tango Control System, its database service will be available on a pre-known network address.
```
hardion: ... so call Tango Host

6/README.md
```diff
@@ -0,0 +1,275 @@
...
+* stores configuration data (properties) used at startup of a device server 
+* acts as nameserver by storing the dynamic network addresses
+* acts as permanent store of dynamic settings which need to be memorised
+* ensures the uniqueness of device name and of aliases
+* keeps list of controlled servers for Starter service
+* limited set functionality provide by file as alternative to network database
+
+
+### General Specification
+The database service shall be implemented as a Tango Device Server.
```
hardion: so call Tango Database (or Database or Database device if you prefer).
You can this term instead of using database service, database device, ...

6/README.md
```diff
@@ -0,0 +1,275 @@
...
+
+In the Tango Control System, its database service will be available on a pre-known network address.
+
+Clients MUST access the database service via TANGO commands requested on the database device.
```
hardion: Client (yet again) is confusing. Imagine the case of Device accessing another Device. It's like explaining the client part of a server...
Better to say Communication MUST be done through Tango Command...

hardion: Tango Command instead of TANGO commands

6/README.md
```diff
@@ -0,0 +1,275 @@
...
+
+As a Tango Device Server, the database SHALL provide State and Status.
```
hardion, line 81: You can leave this. It's already specified in the Device Model RFC.

6/README.md
hardion: Also, note that a Database is not a Tango Device '''Server''' but a Device. (term defined in RFC 2)

6/README.md
```diff
@@ -0,0 +1,275 @@
...
+
+A limited set of the database functionality to support Tango Device Server configuration shall be avaliable as a read-only file.
```
hardion: shall => SHALL
Is the limited set defined further?

6/README.md
```diff
@@ -0,0 +1,275 @@
...
+This is an option for use in the absence of a network service.
+
+The database SHALL enforce uniqueness of device names and aliases.
+
+The device name, the command name, the attribute name, the property name, the device alias name and the device server name SHALL BE case insensitive.
```
hardion: Better to leave this spec to other RFC. Just mention That the Database SHALL enforces the naming convention defined in all Tango RFCs

6/README.md
```diff
@@ -0,0 +1,275 @@
...
+
+The device name, its domain, member and family fields and its alias maximum sizes SHALL BE at least:
```
hardion, line 91: Is it the same for the file case?

6/README.md
```diff
@@ -0,0 +1,275 @@
...
+- DbGetDevicePipeList
+
+### Free Objects
+
+The database SHALL provide a means of querying Free Objects.
+
+The database SHALL implement this free object-related command:
+- DbGetObjectList
+
+### Exported Devices
```
hardion, line 260: This paragraph is probably the most important. 
What do you think to have it first?

6/README.md
```diff
@@ -0,0 +1,275 @@
...
+- domain field: 85
+- family field: 85
+- member field: 85
+- device alias name: 255
+
+### Devices
+The database SHALL provide a means of defining and querying devices.
+
+The database SHALL implement the following device-related commands:
+- DBAddDeviceDbDeleteDevice
```
hardion, line 103: Use a table with the argin and argout type

hardion (commented):
Thanks for this first draft.
As a general comment the text should emphasis the different function of a Database: Registry, Persistance, Browsing.
An explanation of the registry sequence can be given to understand when to use the different Commands from the declaration of a new Device to the export of a new Device.

Otherwise be careful on the term already defined which start with a capital like Tango Server, ...
You can also define ones which are commonly accepted and would reduce confusion when used.
i.e : A Device is said Exported when it is accessible and register in the Database

6/README.md
```diff
@@ -0,0 +1,1078 @@
...
+## Tango Database specification
+
+This specification is intended to formally document the Tango Database. 
+Runtime implementations of the Database in conforming Tango implementations MUST follow this specification.
+
+
+### Goals of the Database
+* provide a means of defining and querying devices.
+* provide means for device servers to be configured from central control
+* serves as nameserver such that such that the Devices are accessible by name
```
lorenzopivetta: serve as nameserver such that the Devices are accessible by name

6/README.md
```diff
@@ -0,0 +1,1078 @@
...
+
+### Features of the Database
+
+* stores configuration data (properties) used at startup of a device server
```
lorenzopivetta: Since we're listing features of... I'd remove the 's' from all the actions, e.g.:
- store configuration...
- act as nameserver...
- act as permanent...
- ...
Also on the other sections.

6/README.md
```diff
@@ -0,0 +1,1078 @@
...
+* define properties for all instances of a class of vaccuum pump
+* check if a electrometer is online or not
+* a user would like to rollback the configuration of a Device when its initialization is failing
+* remove "empty" Devices from Database to clean it up
+* get all Devices of a given class of temperature gauges
+* user wants to see the history (last ten values) of a Property
+* a monitor tool wants the Info of a not running detector Device (exported, host, server)
+* a boot-up process wants to get Starter Level configuration for a list of Devices
+
+## General Specification
```
piogor: I would omit 'General' world.

6/README.md
```diff
@@ -0,0 +1,1078 @@
...
+The database SHALL provide a means of defining and querying devices.
+
+The database service SHALL be implemented as a Tango Device Server.
+
+The Database Device SHALL have no Class Properties or Device Properties.
+
+In the Tango Control System, its database service SHALL be available on a pre-known network address.
+
+Clients and subscribers MUST access the database service via TANGO commands requested on the database device.
```
piogor: I think it is not strong requirement and I would use SHOULD. Various tools in the community are using direct SQL queries.

6/README.md
```diff
@@ -0,0 +1,1078 @@
...
+
+As a Tango Device Server, the database SHALL provide State and Status.
+
+A limited set of the database functionality to support Tango Device Server configuration shall be avaliable as a read-only file.
```
piogor: A limited set of the database functionality to support Tango Device Server configuration SHALL be available as a read-only file. (capital laters in SHALL) and a typo in available.

6/README.md
```diff
@@ -0,0 +1,1078 @@
...
+##### DevicePipeProperty
+- DbGetDevicePipeProperty
+- DbDeleteDevicePipeProperty
+- DbDeleteAllDevicePipeProperty
+- DbPutDevicePipeProperty
+- DbGetDevicePipePropertyHist
+
+
+### Attributes
+The database SHALL provide a means of setting and getting attribute values for a class or device.
```
piogor: 'The database SHALL provide a means of setting and getting information about attributes for a class or device.'
The word `value` suggests that the database is an interface to devices.

6/README.md
```diff
@@ -0,0 +1,1078 @@
...
+- DbGetServerNameList
+- DbRenameServer
+
+##### ServerInfo 
+- DbDeleteServerInfo
+- DbGetServerInfo
+- DbPutServerInfo
+
+### Pipes
+The database SHALL provide a means of querying and deleting pipes.
```
piogor: I suggest '... of managing information about pipes'.

6/README.md
```diff
@@ -0,0 +1,1078 @@
...
+```CLASS/<class name>-><property name>: <property value>```  
+
+CLASS is the keyword to declare a class property definition. Allowed characters after the ’:’ delimiter are space, tabulation or nothing. 
+The ’’ characters around the property value are mandatory due to the ’/’ character contains in the property value.
+
+
+## Related documentation
+[Tango Controls Documentation 9.3.3](https://buildmedia.readthedocs.org/media/pdf/tango-controls/latest/tango-controls.pdf)
+
+[PyTango Documentation Release 8.1.6](http://www.esrf.eu/computing/cs/tango/pytango/v818/PyTango.pdf)
```
piogor: This probably can be updated to the most recent one: https://readthedocs.org/projects/pytango/downloads/pdf/stable/

6/README.md
```diff
@@ -0,0 +1,1078 @@
...
+* remove "empty" Devices from Database to clean it up
+* get all Devices of a given class of temperature gauges
+* user wants to see the history (last ten values) of a Property
+* a monitor tool wants the Info of a not running detector Device (exported, host, server)
+* a boot-up process wants to get Starter Level configuration for a list of Devices
+
+## Specification
+The database SHALL provide a means of defining and querying devices.
+
+The database service SHALL be implemented as a Tango Device Server.
```
gwen-soleil, line 69: * The database service SHALL be implemented w/:
   * a frontend: as Tango Device Server 
   * a backend:  a query and persistency engine (i.e. SQL or NoSQL engine), or a lighter ( i.e. reduced set of  functionalities) engine w/ a single file per device server.

6/README.md
```diff
@@ -0,0 +1,1078 @@
...
+* limited set functionality provide by file as alternative to network database
+
+### Use Cases of the Database
+
+* register a new device server
+* stores the last user setting point of the current provided by a power supply
+* define persistent configuration properties to be used by a camera every time it starts up
+* allow a subscriber or client to get connect to a stepper motor
+* define properties for all instances of a class of vaccuum pump
+* check if a electrometer is online or not
```
gwen-soleil: Not sure to understand this sentence. The Tango database does not check the availability of the control system devices. It just provides the network ID of the devices.

6/README.md
```diff
@@ -0,0 +1,1078 @@
...
+* a user would like to rollback the configuration of a Device when its initialization is failing
+* remove "empty" Devices from Database to clean it up
+* get all Devices of a given class of temperature gauges
+* user wants to see the history (last ten values) of a Property
+* a monitor tool wants the Info of a not running detector Device (exported, host, server)
+* a boot-up process wants to get Starter Level configuration for a list of Devices
+
+## Specification
+The database SHALL provide a means of defining and querying devices.
+
```
gwen-soleil, line 64: * As the Tango database service is a central service of the Tango control system, it SHALL be designed to be highly available and support a high level of client requests (e.g we have peaks of 3500 SQL requests/ s at SOLEIL on the accelerators control system)

6/README.md
```diff
@@ -0,0 +1,1078 @@
...
+* getting, setting and deleting class attribute properties  
+* getting, setting and deleting device attribute properties  
+
+The following set of limitations exist:  
+* no check that the same Device server is running twice
+* no Device or attribute alias name
+* in case of several Device servers running on the same host, the user must manually manage a list of already used network ports
+
+### File syntax
+The file is an ASCII file and SHALL follow a well-defined syntax with predefined keywords.
```
gwen-soleil: The file format is defined in properties RFC.  Shall we move it to this RFC?

6/README.md
```diff
@@ -0,0 +1,1078 @@
...
+
+Generally, comment starts with the ’#’ character and a blank line is ignored.
+
+#### Devices definition.
+DEVICE is the keyword to declare a device(s) definition sequence. The general syntax SHALL be   
+```<DS name>/<inst name>/DEVICE/<Class name>: dev1,dev2,dev3  ```
+Device(s) name can follow on next line if the last line character is ’\’
+
+#### Device property definition. 
+The general device property SHALL be
```
gwen-soleil, line 1042: The naming convention for properties has been already defined in properties RFC-5

6/README.md
gwen-soleil: I agree with Vince.

6/README.md
david-erb: I am not seeing the file format in this PR https://github.com/bourtemb/rfc/tree/properties-rfc-5-2nd-try/5.  Maybe I am looking at the wrong place.

I will leave the File Syntax in the Database RFC just for now, until it is decided.

6/README.md
david-erb: Ok, I can see the confusion, this is device property definition for the file syntax only.  I have changed the heading to reflect this.  Thanks.

6/README.md
bourtemb: Indeed, I removed it from RFC-5 when I rewrote it because this was probably not the best location to put it.
The file syntax was described very briefly in the previous version:
https://github.com/tango-controls/rfc/pull/30/files#diff-192b10cf557bb491182f248eeaf39ae2R181-R188

A very simple example was provided. The syntax was not described using ABNF, which could be a more rigorous way to describe it.

PR #48, opened 2019-10-14
List attribute properties in RFC-4

mliszcz:
This PR addresses the action from the RFC meeting:

 ACTION: Michal will supplement Attribute Model with names of attributes properties

### DIFF
--- a/4/README.md
+++ b/4/README.md
@@ -172,6 +172,38 @@ pointed by *__root_att* MUST be reflected in the forwarded Attribute.
 An Attribute MAY have associated dynamic metadata in form of
 Properties (see [RFC 5/Property](/5)).
 
+Below table summarizes common attribute properties used by Tango.
+Detailed description is provided later in this section.
+
+| Property             | Name in database     | Type   | Default value                                                                                         |
+|----------------------|----------------------|--------|-------------------------------------------------------------------------------------------------------|
+| *description*        | `description`        | string | Not specified / No description                                                                        |
+| *label*              | `label`              | string | Not specified / No label                                                                              |
+| *unit*               | `unit`               | string | Not specified / No unit / *empty*                                                                     |
+| *standard unit*      | `standard_unit`      | number | Not specified / No standard unit                                                                      |
+| *display unit*       | `display_unit`       | number | Not specified / No display unit                                                                       |
+| *format*             | `format`             | string | %6.2f (float),<br/> %d (integer),<br/> %s(string, enum),<br/> Not specified (state, encoded, boolean) |
+| *min value*          | `min_value`          | number | Not specified                                                                                         |
+| *max value*          | `max_value`          | number | Not specified                                                                                         |
+| *min alarm*          | `min_alarm`          | number | Not specified                                                                                         |
+| *max alarm*          | `max_alarm`          | number | Not specified                                                                                         |
+| *min warning*        | `min_warning`        | number | Not specified                                                                                         |
+| *max warning*        | `max_warning`        | number | Not specified                                                                                         |
+| *delta val*          | `delta_val`          | number | Not specified                                                                                         |
+| *delta t*            | `delta_t`            | number | Not specified / 0                                                                                     |
+| *rel change*         | `rel_change`         | number | Not specified                                                                                         |
+| *abs change*         | `abs_change`         | number | Not specified                                                                                         |
+| *archive rel change* | `archive_rel_change` | number | Not specified                                                                                         |
+| *archive abs change* | `archive_abs_change` | number | Not specified                                                                                         |
+| *period*             | `period`             | number | 1000                                                                                                  |
+| *archive period*     | `archive_period`     | number | Not specified                                                                                         |
+| *__value*            | `__value`            | string | Not used yet                                                                                          |
+| *__root_att*         | `__root_att`         | string | Not defined                                                                                           |
+
+> **Note:**
+> Both `delta_val` and `delta_t` MUST be set to non-default values
+> in order for RDS (*read different than set*) alarms to work.
+
 General properties:
 * *description*, providing textual information about the Attribute,
 * *label*, providing textual label to use as Attribute's name,


mliszcz:
@srgblnch @bourtemb please check as you did the review of the original PR. I cannot add you as reviewers due to lack of permissions.

4/README.md
```diff
@@ -172,6 +172,32 @@ pointed by *__root_att* MUST be reflected in the forwarded Attribute.
 An Attribute MAY have associated dynamic metadata in form of
 Properties (see [RFC 5/Property](/5)).
 
+Below table summarizes summarizes all attribute properties.
```
bourtemb: Typo: => Below table summarizes all attribute properties.

This table does not list all attribute properties.
It lists all the attribute properties common to all Tango attributes.
The device server programmer can define his own attribute properties so in theory there could be an infinite number of attribute properties.
"__value" and "__root_att" attribute properties are not listed in this table and are also considered as attribute properties but "__value" is used only for memorized attributes and "__root_att" is used only for forwarded attributes.

4/README.md
```diff
@@ -172,6 +172,32 @@ pointed by *__root_att* MUST be reflected in the forwarded Attribute.
...
+| *max warning*        | `max_warning`        | number | Not specified                                                                                         |
+| *delta val*          | `delta_val`          | number | Not specified                                                                                         |
+| *delta t*            | `delta_t`            | number | Not specified                                                                                         |
+| *rel change*         | `rel_change`         | number | Not specified                                                                                         |
+| *abs change*         | `abs_change`         | number | Not specified                                                                                         |
+| *archive rel change* | `archive_rel_change` | number | Not specified                                                                                         |
+| *archive abs change* | `archive_abs_chagne` | number | Not specified                                                                                         |
```
bourtemb: archive_abs_chagne => archive_abs_change

4/README.md
```diff
@@ -172,6 +172,32 @@ pointed by *__root_att* MUST be reflected in the forwarded Attribute.
...
 Properties (see [RFC 5/Property](/5)).
 
+Below table summarizes summarizes all attribute properties.
+Detailed description is provided later in this section.
+
+| Property             | Name in database     | Type   | Default value                                                                                         |
+|----------------------|----------------------|--------|-------------------------------------------------------------------------------------------------------|
+| *description*        | `description`        | string | Not specified / No description                                                                        |
+| *label*              | `label`              | string | Not specified / No label                                                                              |
+| *unit*               | `unit`               | string | Not specified / No unit                                                                               |
```
bourtemb: The default unit property C++ library default has been changed from "No Unit" to empty string in commit https://github.com/tango-controls/cppTango/commit/4aac0cc

4/README.md
```diff
@@ -172,6 +172,32 @@ pointed by *__root_att* MUST be reflected in the forwarded Attribute.
...
+| *max value*          | `max_value`          | number | Not specified                                                                                         |
+| *min alarm*          | `min_alarm`          | number | Not specified                                                                                         |
+| *max alarm*          | `max_alarm`          | number | Not specified                                                                                         |
+| *min warning*        | `min_warning`        | number | Not specified                                                                                         |
+| *max warning*        | `max_warning`        | number | Not specified                                                                                         |
+| *delta val*          | `delta_val`          | number | Not specified                                                                                         |
+| *delta t*            | `delta_t`            | number | Not specified                                                                                         |
```
bourtemb: delta_t default value seems to be 0 in some parts of the code:
https://github.com/tango-controls/cppTango/blob/tango-9-lts/cppapi/server/multiattribute.cpp#L67

and seems to be set to "Not specified" when someone tries to get attribute properties  from Attribute::get_properties(Tango::AttributeConfig_3 &conf) method:
https://github.com/tango-controls/cppTango/blob/tango-9-lts/cppapi/server/attrgetsetprop.cpp#L183-L184

bourtemb (commented):
Thanks for the commit!
Please see my comments (2 typos, comment on some default values and special attribute properties)

4/README.md
mliszcz: The behavior I observe is:

* When I put some number  in Jive in delta_t (e.g. "2"):
    ```
    >>> tango.AttributeProxy("sys/tg_test/1/double_scalar").get_property("delta_t")
    {'delta_t': ['2']}
    ```
* When I put "0", it changes to `None` in Jive but pytango API still returns old value (2)

* When I put "Not specified", it also changes to `None` in Jive but still pytango returns old value.

So it looks like neither is working. What do you propose to put in the table?

bourtemb: > The behavior I observe is:
> 
>     * When I put some number  in Jive in delta_t (e.g. "2"):
>       ```
>       >>> tango.AttributeProxy("sys/tg_test/1/double_scalar").get_property("delta_t")
>       {'delta_t': ['2']}
>       ```
> 
>     * When I put "0", it changes to `None` in Jive but pytango API still returns old value (2)
> 
>     * When I put "Not specified", it also changes to `None` in Jive but still pytango returns old value.
> 
> 
> So it looks like neither is working. What do you propose to put in the table?

The behaviour you are observing seems weird indeed but I think it is due to several factors.
If you set delta_t to a numeric value and if delta_val is not set, then the new delta_t value will not be taken into account until delta_val **and** delta_t are set.
If you set only one of the 2, you will see the following log message on the server side:
`WARN test/rb/tg_test-1 RDS (Read Different Set) incoherent in attribute double_scalar (only delta_t is set)`

Then, I think there is something which does not seem to be handled properly by jive when resetting delta_t and delta_val attribute property values individually in this use case.
If you use right-click and "Reset to default value" or "Return to xxx default value", jive will do something which will remove the info from the DB and then trigger a restart of the device and the attribute property will be back to the correct default value, even when getting it from PyTango.

It is clear that there seems to be a difference of behaviour between JTango (or jive?) and PyTango when reading these delta_t and delta_val attribute properties (at least).

bourtemb: > It is clear that there seems to be a difference of behaviour between JTango (or jive?) and PyTango when reading these delta_t and delta_val attribute properties (at least).

After some tests, it looks like the PyTango commands you were using are only looking at the values which are stored in the database whereas jive seems to ask to the device.

If you have a device server where delta_t and delta_val are both defined in the code (by POGO) and if the user didn't overload these attribute properties, the PyTango commands you were using are returning empty values whereas jive is displaying the values which are defined by code.

bourtemb: I think the device will store delta_t and delta_val attribute properties in the database only if both are defined (different than lib default value) and at least one of them is not equal to the default value defined by code (class/code/lib).
In the same way, they are deleted from the DB only if both are equal to code default value.

If you define: delta_t =2 and delta_val = 0.5 from POGO and don't overload these attribute properties and then do a "Reset to lib default value" from jive, then you will see:
delta_t = 0 and delta_val = 'Not specified' in the DB.

The behaviour seems a bit complex to decide when to delete these attribute properties from the DB. It looks like they are deleted when their value is equal to the value they get from the code.
And for delta_t and delta_val it seems to be even more complex because these 2 properties are linked together.

mliszcz: Hi Reynald,
I propose to add below note then. Is it ok for you?

> Note: Both `delta_val` and `delta_t` MUST be set to non-default values in order for RDS (*read different than set*) alarms to work.

bourtemb: Hi @mliszcz . On the principle I agree.
I am not sure this sentence is in good English though... (in order for ...) I would reformulate it like that but if a native English speaker is around, he could probably propose something better:

> RDS (Read Different than Set) alarms work only if `delta_val` and `delta_t` are both set to non-default values.

bourtemb: > Note: Both `delta_val` and `delta_t` MUST be set to non-default values in order for RDS (_read different than set_) alarms to work.

This sentence seems actually correct... (https://www.merriam-webster.com/dictionary/in%20order%20for%20(someone%20or%20something)%20to)... So forget my previous comment and please forgive my low level in English language... ;-)

PR #49, opened 2019-10-14
Assing RFC number to: Logging service (RFC-14)

mliszcz:
Fixes action from 30-09 meeting:

ACTION: Michal: update table on the wiki and in the README with a new RFC number: RFC-14 Logging Service. David will be the author.


piogor (approved):
Thanks @mliszcz

PR #50, opened 2019-10-14
Rename RFC-9 to: Data types

mliszcz:
Action from 30-09 meeting:

 Rename RFC-9 to Type (from previous meeting), ACTION: Michal: rename the RFC on the wiki and in the README.


piogor (approved):
Thanks, @mliszcz.

PR #52, opened 2019-10-26
WIP Logging (RFC-14)

david-erb:


6/README.md
```diff
@@ -0,0 +1,1078 @@
+---
```
piogor: Hi David, this file, probably, should not be included in this PR :-).

Ingvord:
@david-erb-maxiv thanks for the contribution!

I have looked through the RFC and here are my thoughts:

Most of it looks like a specification for a logging framework e.g. log4j or similar. Which is maybe fine as Tango 9 embeds logging framework (log4tango in case of C++ and log4j in case of Java). It is historical decision so lets live with that.

However I would personally omit specification of the logging framework already in this RFC e.g. file sizes etc. First of all it implies some restrictions which may not be reasonable in some cases e.g. one uses external logging framework which not fully compatible with this RFC while being much more flexible and powerful or solves specific use case for a particular installation. This especially true for Java and maybe for Python as well.

Following the discussion at RFC meeting on 28/10/2019 (@piogor which is missing from the minutes BTW) I am totally agree with @bourtemb that RFC for Tango V9 must be way more simple, otherwise we all losing the main goal - freeze existing features, move forward with V10

This is a general comment. I also put some inline comments.

14/README.md
```diff
@@ -0,0 +1,173 @@
...
+- debug the software
+- report on errors
+- give regular information to user
+- provide timely information of the activities of the Device
+- report the version of the softare
+- provide clues to the internal changes happening in the Device
+
+At run-time, the TLS supports such activities as:
+- changing logging verbosity and destination of any Tango Device
+- capture logs to files
```
Ingvord, line 48: This and all below are implementation specific, usually depend on the 3rd party library used for logging

14/README.md
```diff
@@ -0,0 +1,173 @@
...
+- display log entries on the command line console
+- interactively view log entries as they are produced by a remote Tango Device
+- review logs already written to files
+- make graphical dashboards with such tools as Kibana and Grafana
+
+## Specification
+
+### Logging Properties 
+TLS SHALL use four Device Properties to control Device Server logging at startup.  The Properties may be set during runtime in the normal way.
```
Ingvord: Would be nice to name the properties

Ingvord, line 57: Should all of this reads MUST instead of SHALL, should not? In my understanding SHALL means desirable but not necessary, while MUST is simply must otherwise not compliant to the spec

14/README.md
```diff
@@ -0,0 +1,173 @@
...
+SHALL give the name of one or more Logging Target (see below), separated by commas.
+
+#### logging_rft property
+SHALL cause the maximum file size written to be the given value.  If a file grows past this size, SHALL be closed, renamed and a new file opened.
+
+#### logging_path property
+SHALL override the TANGO_LOG_PATH environment variable and SHALL only apply to classes derived from the DServer class.
+The value SHALL specify where log files are written for Logging Target "file" (with no filename).
+
+### Logging Targets
```
Ingvord, line 72: This section is implementation specific and as everything reads SHALL which is my understanding desirable but not required I would move this into a dedicated section called, for instance, "Tango V9 Implementation"

14/README.md
```diff
@@ -0,0 +1,173 @@
...
+logging with print: ```print >>self.debug_info, "read voltage attribute"```
+
+logging with decorators: ```@PyTango.DebugIt()```
+
+The examples just given are for the DEBUG log level.  There SHALL BE equivalent macros for the other log levels, except OFF.
+
+
+### Java API
+TLS SHALL provide configuration to allow Java to log messages in the following manner:   
+```get_logger().debug(Initializing device  + get_name());```
```
Ingvord, line 129: As Java developer I would say 99,9999% of the applications use slf4j in which logger initialization and the command above looks different. So let's just keep this aside

In fact the whole API section may be replaced with "TLS MUST provide convenient API for clients to log messages" and then either move all above to Tango V9 implementation specific section or just put a link to API docs

14/README.md
```diff
@@ -0,0 +1,173 @@
...
+The value SHALL specify where log files are written for Logging Target "file" (with no filename).
+
+### Logging Targets
+TLS SHALL allow Logging Targets, that is, destinations, for messages.  A Device Server SHALL emit messages to all targets (comma-separated) to its logging_target Property.
+#### console
+output SHALL be delivered to stdout
+#### file
+output SHALL be written to a standard location in a directory specified by TANGO_LOG_PATH if nonblank, otherwise in a directory created under /tmp/tango-*.
+#### file::filename
+output SHALL be written to the filename given.  File format SHALL be log4j compatible.
```
Ingvord, line 79: Well, I would say "TLS MUST inform client about the log message format" rather than bind ourselfs to log4j. 

Guess we will do so in Tango V10

14/README.md
```diff
@@ -0,0 +1,173 @@
...
+
+printf-like: ```LOG_DEBUG(("Msg#%d - Hello world", i++));```
+
+stream-like: ```DEBUG_STREAM << "Msg#" << i++ << "- Hello world" << endl;```
+
+These two logging requests are equivalent. Note the double parenthesis in the printf version.
+
+The examples just given are for the DEBUG log level.  There SHALL BE equivalent macros for the other log levels, except OFF.
+
+TLS SHALL provide a means to log in the name of a device.  This SHALL be provided when the Device must is a member of a class inheriting from the ```Tango::LogAdapter``` class.
```
Ingvord, line 113: This is unclear to me

14/README.md
```diff
@@ -0,0 +1,173 @@
...
+The examples just given are for the DEBUG log level.  There SHALL BE equivalent macros for the other log levels, except OFF.
+
+
+### Java API
+TLS SHALL provide configuration to allow Java to log messages in the following manner:   
+```get_logger().debug(Initializing device  + get_name());```   
+The example just given is for the DEBUG log level.  There SHALL BE equivalent macros for the other log levels, except OFF.
+
+
+### Log Consumer
```
Ingvord, line 133: This is good written section

14/README.md
```diff
@@ -0,0 +1,173 @@
...
+When a Device Target is specified as "device", this SHALL designate a Device Server which is implementing the Log Consumer interface.
```
Ingvord: Again I would replace SHALL with MUST

14/README.md
```diff
@@ -0,0 +1,173 @@
...
+The array of strings composing the argument SHALL be interpreted with the following meanings:
+- details[0] : the timestamp in millisecond since epoch (01.01.1970)
+- details[1] : the log level
+- details[2] : the log source (i.e. device name)
+- details[3] : the log message
+- details[4] : the log NDC (contextual info) - Not used but reserved
+- details[5] : the thread identifier (i.e. the thread from which the log request comes from)
+
+
+### GUI for viewing logs
```
Ingvord, line 147: I would remove this, as GUI is not part of the kernel library as any 3rd party app or library

Or just say "TLS client MAY be a GUI app. Tango V9 distribution has LogViewer app" and put a link to LogViewer repo or docs

14/README.md
```diff
@@ -0,0 +1,173 @@
...
+LogViewer SHALL have two modes: static and dynamic.   
+
+- static: LogViewer SHALL be started with the name of the Log Consumer Device name and it SHALL receive all messages from Devices who have this name as their target.
+
+- dynamic: LogViewer SHALL let the user use the GUI to specify which devices LogViewer will monitor.
+
+
+## References
+The TANGO Control System: Reference Section A.1.5: The device logging
+http://www.esrf.eu/computing/cs/tango/tango_doc/kernel_doc/ds_prog/node11.html#SECTION001115000000000000000
```
Ingvord: All links must point to https://tango-controls.readthedocs.io i.e. official Tango documentation not ESRF

david-erb: Agreed, and done.

PR #53, opened 2019-11-05
WIP RFC-15 Dynamic attribute and command

HelgaMond:
This RFC covers dynamic attribute and command features in Tango V9.


15/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+
+## Goals
+
+This specification describes the Tango kernel dynamic Attribute and Command API. 
+
+
+## Use Cases
+
+The primary use case is to allow server client code to create Attributes and/or Commands in runtime.
```
bourtemb: I find "server client code" a bit confusing here even though it is explained in the diagram below and in the definitions section, which is below too.
I could suggest the following as alternative (other proposals are welcome of course):

> `The primary use case is to allow Tango Device Server programmers to create Attributes and/or Commands in runtime.`

15/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+Reasons for that:
+- create Attributes/Commands based on underlying library;
+- create generic Tango device classes which interfaces entirely/partly unknown at the moment of creation.
```
bourtemb: which => with
or
create generic Tango device classes which interfaces **are** entirely/partly unknown at the moment of creation
?

15/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+Consider a Tango device class that represents a serie of detectors. Each detector differs a bit in terms of supported features, while the most part of the code could be the same for each detector. These small differentiations may be implemented as dynamic Attributes or Commands.
```
bourtemb: Good example!

I would suggest to add other use cases:

Consider a Tango Device Class that represents a residual gaz analyzer. The device will provide measurement results for many different masses. The results can be presented as a Tango spectrum attribute showing the measurements for all the masses but it can be also very interesting to create an attribute for each different mass, then the user could monitor this specific mass or define some alarm thresholds or configure archiving events for this specific mass. Dynamic attributes may be used to create easily these numerous attributes whose behaviour is almost identical.

Consider a Tango Device my/simu/device simulating the interface of another Tango Device my/original/device, for tests purposes. my/simu/device could query the interface of my/original/device at runtime and create dynamically the same attributes and commands as my/original/device. The simulated device could then be used to test high level client applications which would then talk to the simulated device instead of the original device.

15/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+This API MAY accept configuration for each Attribute or/and Command. 
+
+Accepted configuration MUST be the same as described in RFC-3 (The command model) and RFC-4 (The attribute model)
+
+As the result of the call, API creates corresponding Attribute/Command fully compliant to RFC-4/RFC-3.
+
+Implementation of the API MAY send an Event (INTERFACE_CHANGE, see RFC-12) in response to a new Attribute/Command creation.
+
+
```
bourtemb: What about attribute and command removal?

15/README.md
```diff
@@ -0,0 +1,67 @@
+---
+domain: rfc.tango-controls.org 
+shortname: 15/Dynamic-attr-cmd 
+name: The dynamic attribute and command 
+status: raw 
+editor:
```
bourtemb:
```suggestion
editor:  Reynald Bourtembourg `<at esrf.fr - reynald.bourtembourg>`
```

bourtemb (commented):
Thank you @HelgaMond and @ingvord for the good work done on this RFC.

15/README.md
bourtemb: I think something like: 

> Tango kernel library MUST provide API to server client to remove dynamic Attributes, Commands.

is missing.

Do we create another section dedicated to the removal?

> ## API that allows to remove Attributes and Commands in runtime

Or do we change the title:

> ## API that allows to create Attributes and Commands in runtime

into 
> ## API that allows to create and remove Attributes and Commands in runtime

15/README.md
```diff
@@ -0,0 +1,70 @@
...
+Accepted configuration MUST be the same as described in RFC-3 (The command model) and RFC-4 (The attribute model)
+
+As the result of the call, API creates corresponding Attribute/Command fully compliant to RFC-4/RFC-3.
+
+Implementation of the API MAY send an Event (INTERFACE_CHANGE, see RFC-12) in response to a new Attribute/Command creation/removal.
+
+
+
+> In Tango V10
+All Commands and Attributes MUST be defined as dynamic i.e. use this API
```
bourtemb: I would be less radical and propose to replace MUST with MAY.

15/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+Tango kernel library MUST provide API to server client to add new Attributes, Commands.
+
+This API MAY accept configuration for each Attribute or/and Command. 
+
+Accepted configuration MUST be the same as described in RFC-3 (The command model) and RFC-4 (The attribute model)
+
+As the result of the call, API creates corresponding Attribute/Command fully compliant to RFC-4/RFC-3.
+
+Implementation of the API MAY send an Event (INTERFACE_CHANGE, see RFC-12) in response to a new Attribute/Command creation/removal.
```
bourtemb: Just a general question to the community and RFC experts:
Do we consider that a valid Tango implementation is allowed to NOT implement the INTERFACE_CHANGE events as the MAY verb is suggesting here?

bourtemb:
There is a conflict on the README.md file.
Could you please solve it?

15/README.md
HelgaMond: Any "static" can be defined via "dynamic" in implementation. Therefore there is a "MUST" word to enforce to use "dynamic".

HelgaMond:
@bourtemb Thank you for the comments! The conflicts have been resolved.

bourtemb:
> @bourtemb Thank you for the comments! The conflicts have been resolved.

It seems like some changes in the root README.md have been lost during this conflicts resolution.
For instance, the name of the README files in the RFC directories have been renamed from 2/README.md to 2/Device.md, etc...

According to that, I think 15/README.md file in the pull request should also be renamed.

bourtemb (approved):
Thanks @HelgaMond 
I still don't like much the note on V10 which is imposing a way to implement the features but since this is V9 specifications and not V10 specifications, I guess this is not that relevant here and I can live with that.

So, it looks good to me.
Maybe we should update the copyright year? But I guess that's not that important.
If someone else wants to review, I will wait a bit before merging.

bourtemb (dismissed):
Ooops.
The last changes I asked were not yet implemented.
I'll wait for these changes...
> It seems like some changes in the root README.md have been lost during this conflicts resolution.
> For instance, the name of the README files in the RFC directories have been renamed from 2/README.md to 2/Device.md, etc...
> 
> According to that, I think 15/README.md file in the pull request should also be renamed.

HelgaMond:
Dear @bourtemb Thank you for the comments! Could you please check if the problems with the resolution are solved? Thank you!

bourtemb (commented):
Thanks @HelgaMond !
Looks good to me.
I guess it would be convenient to add links to the other related RFCs too.
Let's wait for @piogor 's answer on the convention for the links.
I also noticed some inconsistencies when using Upper case letters as first letter of some words (Tango Device Class vs Tango device class for instance)

piogor:
> Let's wait for @piogor 's answer on the convention for the links.

Hi @bourtemb and @HelgaMond, I didn't realise the question, however, the convention is like the following: `[RFC 2/Device](/2)`
Is it what you are asking for?

bourtemb:
> Hi @bourtemb and @HelgaMond, I didn't realise the question, however, the convention is like the following: `[RFC 2/Device](/2)`
> Is it what you are asking for?

Yes, thanks for the answer @piogor . So, it is not like in the main Readme.md file where we link to the md file directly?
[RFC-2](2/Device.md)

piogor:
> Yes, thanks for the answer @piogor . So, it is not like in the main Readme.md file where we link to the md file directly?
> [RFC-2](2/Device.md)

We may need to synchronise the link convention in all the RFCs, later. However, for the moment, let's stay like this.

piogor:
Added this: #88

bourtemb (commented):
Thanks @HelgaMond  
As for the Pipe RFC, the content looks good to me.
There is still an outstanding work to be done to add links to other RFCs but this can be done later on when a decision will have been taken in https://github.com/tango-controls/rfc/issues/88.
The question of when using upper case first letters can also be addressed later on when a global harmonization work on all the RFCs will be done.
So, if no objection, I will merge this PR this afternoon after the RFC meeting.

HelgaMond:
@bourtemb Thank you very much! Sure, please, merge the RFC.

PR #58, opened 2019-11-07
WIP Device server model (RFC 8)

hardion:
Here is a first version of the Device Server spec.
They are still work to be done:
- [x]  unexport
- [x] spelling and formatting


8/README.md
```diff
@@ -0,0 +1,139 @@
...
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119).
+
+
+## Tango Device Server specification
+
+ The specication of '''Device Server''' aims the process of Device management, being it self a Device.
+
+### Goals
+
+ A Device Server in a Tango Controls System aims to control the process of communication and lifecycle of one or several Device.
```
lorenzopivetta: Suggest to rephrase as "In a Tango Control System a Device Server is in charge of managing the life cycle and the communication process of one or several Tango Devices.

8/README.md
```diff
@@ -0,0 +1,139 @@
...
+
+Additionally, it aims to:
+
+* Provide a blackbox that register the operation executed on every Device
```
lorenzopivetta: s/register/registers/
s/operation/operations/

8/README.md
```diff
@@ -0,0 +1,139 @@
...
+* to manage several Device in the same process
+
+* to share the same channel of communication
+
+* to serialise the operation per Device
+
+
+## Specification
+
+The Device Server manage the communication with one or several Device.  Additionally it enforces the behaviour compatibility with the Tango Control System and offer different meta service.
```
lorenzopivetta: s/manage/manages/
s/behaviour compatibility/compatibility of the behavior/

8/README.md
```diff
@@ -0,0 +1,139 @@
...
+
+The Device Server SHALL follow these specification below in order to  control and monitor this communication:
```
lorenzopivetta: Suggest to rephrase "The Device Server SHALL implement the following specifications to control and monitor the Device:"

8/README.md
```diff
@@ -0,0 +1,139 @@
...
+
+* A Device Server represents/interfaces the process which manage one or several Device
```
lorenzopivetta: I'll remove "/interfaces"

8/README.md
```diff
@@ -0,0 +1,139 @@
...
+
+* A Device Server is a also a Device itself.
```
lorenzopivetta: Suggest to rephrase "A Device Server ia a Tango Device itself"

8/README.md
```diff
@@ -0,0 +1,139 @@
...
+
+The specification introduces as well the definition of the process which are used for the Device Server Name:
+
+* A Server is a type of process managed by a Device Server
+
+* An Instance is an unique identifier as several processes of the same Server MAY exist in a Tango Control System
+
+* A Server Instance is finally what identify the actual process in the Tango Control Sytem
+
+* A Server Instance CAN be interfaced by only one Device Server
```
lorenzopivetta: s/interfaced/managed/

8/README.md
```diff
@@ -0,0 +1,139 @@
...
+
+
+Additionally:
+* A Device Server MAY be call DServer or Admin Device but the terminology Device Server is RECOMMENDED
```
lorenzopivetta: Well, Administration Device is quite common use, why do you suggest to drop it?

8/README.md
hardion: Sounds good !

8/README.md
hardion: DServer and Admin device  are used for the same meaning in different facilities. But for reference we can recommend one. I choose the one from the RFC title.

lorenzopivetta: Sounds good

bourtemb: A device server MAY be called Admin Device?
I never heard this one.
For me, the admin device is a _Device_ (not a _Device Server_).
For me, an admin device has a device name like dserver/server_exec_name/instance_name

hardion: @bourtemb good comment! Effectively Device Server are called from different name depending on the facility. But the good news is we are all right. In this RFC I wrote a Device Server is Tango Device it self and naming convention is dserver/server/instance

bourtemb: OK, so we definitely are using different names here in different institutes.
For me a device server is something different than an admin device.
I would not write:
"A Device Server SHALL be a Tango Device itself"
because this will bring confusion and encourage users to use the words "Device Server" and "Tango Device" for the same thing. They are not the same things. A device server is exporting and managing some devices. It is not the same thing. The admin device is a special device which SHALL be exported by any Tango Device Server and which is there to provide an interface to interact with the Device Server.
Instead of "A Device Server SHALL be a Tango Device itself", I would rather write something like:
"A Device Server SHALL export at least an administration Device"

lorenzopivetta: This is very similar to what I wrote in the PR #79, last line:
- The Device Server SHALL have a management device

Do we want to allow for both the forms "administration" and "management" Device?

hardion: I see your point. If I understood you, '''Device Server''' is the system process and the '''Admin Device''' is the Tango Device which connected to the process '''Device Server'''.
This is very confusing considering that the Class is DServer (BTW missing in the RFC).
For me it's OK to call '''Device Server''' either the process of managing device and the dserver/xxx/yyy Device. At the end this is the role of this Tango Device to manage other devices.

hardion: I think the people were too lazy to call it the "Device Server Device" like we would say for the "PowerSupply Device"

bourtemb: > I see your point. If I understood you, '''Device Server''' is the system process and the '''Admin Device''' is the Tango Device which connected to the process '''Device Server'''.

Yes.

> This is very confusing considering that the Class is DServer (BTW missing in the RFC).

I understand. So the confusion is coming from the admin device class name.
The admin device is a device of class DServer. It is an object representing/controlling the Device Server, so it was quite natural to name this class like that (or DeviceServer).
When you have a Tango class controlling a PowerSupply, you naturally name this class _PowerSupply_. Same principle, here.

> For me it's OK to call '''Device Server''' either the process of managing device and the dserver/xxx/yyy Device. 

I would personally try to avoid to call a Device, a Device Server. _C'est un abus de langage_, as we would say in French. We should not encourage misleading terms in the specifications I think.

> At the end this is the role of this Tango Device to manage other devices.

The role of the admin device is to manage the Device Server, so part of it is to manage other devices, yes.

bourtemb: > This is very similar to what I wrote in the PR #79, last line:
> 
>     * The Device Server SHALL have a management device
> 
> 
> Do we want to allow for both the forms "administration" and "management" Device?

I guess what is important is the role of the device here. So we can put what we want.
But if we want to relate more easily to the current documentation and tools, I think it would be better to use the term _administration device_.

hardion: > The role of the admin device is to manage the Device Server, so part of it is to manage other devices, yes.
Is it good to say :
 the Device Server for the role of managing devices,
 the Device Server SHALL be implemented as a Tango Device, which is called the Admin Device
 the Admin Device should follow this naming convention ... dserver/server/instance

bourtemb: Maybe this could help:
Extract from the [Tango-Controls documentation's glossary](https://tango-controls.readthedocs.io/en/latest/reference/glossary.html):

> **device server**
    A Device Server is a program (executable) which is able to create devices of certain classes. A Device Server may implement one or multiple classes and instantiate one or more devices. A running device server is called a Device Server instance.
>
> **instance**
**device server instance**
    A running device server is called a device server instance. So it means, it is a process. Every device server instance has an unique name in Tango Controls by which it can be referenced. The name is built as {DeviceServerName}/{instanceName}. For each running device server the system creates a special device of DServer class: dserver/{DeviceServerName}/{instanceName}. This device provides a management facility for the corresponding device server instance (see DServer class device commands).

This page might help as well?:
https://tango-controls.readthedocs.io/en/latest/development/advanced/TangoDeviceServerModel.html

8/README.md
```diff
@@ -0,0 +1,164 @@
...
+
+This Specification is a [free and open standard](http://www.digistan.org/open-standard:definition) and is governed by the Digital Standards Organization's [Consensus-Oriented Specification System](http://www.digistan.org/spec:1/COSS).
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119).
+
+
+## Tango Device Server specification
+
+ This spefication formally document the Tango Device Server model.
```
gwen-soleil:
```suggestion
 This specification formally document the Tango Device Server model. 
```

8/README.md
```diff
@@ -0,0 +1,164 @@
...
+
+* Provide a blackbox that registers the operations executed on every Device
+
+* Be usable as a Device
+
+### Use Cases
+
+The main use cases for Device Server are:
+
+* to manage several Device in the same process.
```
gwen-soleil:
```suggestion
* to manage several Devices in the same process. 
```

8/README.md
```diff
@@ -0,0 +1,164 @@
...
+
+* to manage several Device in the same process. 
+
+* to share the same channel of communication
+
+* to provide the connection protocols for accessing the managed Device
+
+* to provide different services like logging
+
+* to serialise the operation per Device
```
gwen-soleil:
```suggestion
* to serialize the operations per Device
```

8/README.md
```diff
@@ -0,0 +1,164 @@
...
+
+* to sign up or sign out to the central registry 
+
+
+
+## Specification
+
+The Device Server manages the communication with one or several Device.  Additionally it enforces the compatibility of behaviour with the Tango Control System and offer different meta service.
```
gwen-soleil:
```suggestion
The Device Server manages the communication with one or several Devices.  Additionally it enforces the compatibility of behaviour with the Tango Control System and offer different meta service.
```

8/README.md
```diff
@@ -0,0 +1,164 @@
...
+
+The Device Server SHALL implement the following specifications to control and monitor the Device:
+
+* A Device Server represents the process which manage one or several Device
```
gwen-soleil:
```suggestion
* A Device Server represents the process which manage one or several Devices
```

8/README.md
```diff
@@ -0,0 +1,164 @@
...
+
+* A Device Server SHALL managed at least one Device.
```
gwen-soleil:
```suggestion
* A Device Server SHALL manage at least one Device. 
```

8/README.md
bourtemb:
```suggestion
* A Device Server represents the process which manages one or several Devices
```

8/README.md
gwen-soleil: My comments:
I have always used the same definitions of the Tango documentation. So one device of class "DServer" is a device started automatically by the "Device Server". Maybe a specific RFC should be created for the "DServer"? But the terminology distinction between DServer and 'Device Server' have always been confusing at SOLEIL, especially for new comers.  So "AdminDevice" seems to be more explicit for me so I have used that name in JTango.

8/README.md
hardion: s -> british
z -> us
I see your preference ;-)

gwen-soleil: oops, sorry. You can keep the british version ;)

8/README.md
hardion: This paragraph has been rewritten according to your comments. @bourtemb @lorenzopivetta @gwen-soleil Can you review this paragraph?

8/README.md
```diff
@@ -0,0 +1,166 @@
...
+
+* to manage several Devices in the same process. 
+
+* to share the same channel of communication
+
+* to provide the connection protocols for accessing the managed Device
+
+* to provide different services like logging
+
+* to serialise the operation per Device
```
bourtemb:
```suggestion
* to serialise the operations per Device
```
I agree with @gwen-soleil and I would put an 's' at the end of operation here

8/README.md
```diff
@@ -0,0 +1,166 @@
...
+
+* to sign up or sign out to the central registry 
+
+
+
+## Specification
+
+The Device Server manages the communication with one or several Devices.  Additionally it enforces the compatibility of behaviour with the Tango Control System and offer different meta service.
```
bourtemb:
```suggestion
The Device Server manages the communication with one or several Devices.  Additionally it enforces the compatibility of behaviour with the Tango Control System and offers different meta services.
```

or

```suggestion
The Device Server manages the communication with one or several Devices.  Additionally it enforces the compatibility of behaviour with the Tango Control System and offers a different meta service.
```

8/README.md
```diff
@@ -0,0 +1,166 @@
...
+* A Device Server MAY manage Devices belonging to different Classes
+
+* A Device MUST NOT be managed by more than one Device Server
+
+* A Device Server instance reference SHALL be unique in a Tango Control system
+
+* The communication to the Device Server SHALL be done through a Tango Device so called Administration Device or Admin Device .
+
+
+The specification introduces as well the definition of the process which are used for the Admnistration Device Name:
```
bourtemb:
```suggestion
The specification introduces as well the definition of the processes which are used for the Administration Device Name:
```

bourtemb: Or did you mean:
```suggestion
The specification introduces as well the definition of the process used to build the Administration Device Name:
```
?

8/README.md
```diff
@@ -0,0 +1,166 @@
...
+```ABNF
+administration-device-name = dserver "/" server "/" instance
+dserver = "dserver" ; dserver is a reserved domain following the device-name convention.
+server = family
+instance = member
+```
+
+* The member part of the Administration Device Name SHALL be the Instance
+
+* The family part of the Administration Device Name SHALL be the Server name
```
bourtemb: I suggest to reorder these 2 lines to mention the family part before the member part.

8/README.md
bourtemb: By the way, the serialisation can be per Device, per Class or per Process too

8/README.md
```diff
@@ -0,0 +1,166 @@
...
+
+#### Device Server interface
+
+
+| Property Name | description |
+|---|---|
+| polling_threads_pool_conf | TODO |
+
+| Command Name | description |
+|---|---|
```
bourtemb: This table seems to expect 3 columns.
If you want to add a description too, it should be with 4 columns (WIP I guess)

8/README.md
```diff
@@ -0,0 +1,166 @@
...
+
+The phase called Device Unexport is intended to prepare the all Device for stopping operate.
+
+* The Device Server MUST take in charge of the Device Unexport
+
+* The Device Unexport SHALL gracefully stop the Device operation, even if there are requests in progress.
+
+* The Device Unexport SHALL gracefully stop the communication with any Tango Client 
+
+* The Device Unexport SHOULD unregister from the Tango Control system by executing the Device Sign Out of the Database([RFC-6]()), in order to notify that all managed Device are not accessible
```
bourtemb:
```suggestion
* The Device Unexport SHOULD unregister from the Tango Control system by executing the Device Sign Out of the Database([RFC-6]()), in order to notify that all managed Device are no longer accessible
```

8/README.md
hardion: True. This is the use case section, I will add it after to my branch.

hardion:
@lorenzopivetta Ready for Draft

lorenzopivetta:
Thanks @hardion  . One question, also for @piogor and @bourtemb : is it bettere to merge each single commit independently or squash into a single one? In principle I do not like squash commits... but...

bourtemb:
@lorenzopivetta , it's a matter of taste.
I personally prefer the squash commits method because I then find the history cleaner and lighter... 
But some other persons do not like it because it "changes" the history.
I don't have a strong opinion on this matter, just a preference.
In any case I think it is important to keep the Pull Request number in the commit log so people can refer to it easily if needed (if github is still there) .

lorenzopivetta:
Thanks @bourtemb. Indeed, in this case a squash may be cleaner. And we're writing the beginning... (of the history)

PR #59, opened 2019-11-07
WIP RFC-5 (Property) - 2nd attempt

bourtemb:
Follow-up of #30

### DIFF
--- a/5/README.md
+++ b/5/README.md
@@ -35,33 +35,33 @@ Runtime representation of Property in conforming Tango implementations MUST foll
 
 A Property represents a configuration parameter in Tango.
 
-There are 4 kinds of Tango Properties:
+### Use Cases
+
+Here are some Property use cases:
+
+* To connect a real network equipment a Property can define the target IP address or hostname
+* To change the representation of an information in order to be more user friendly a Property can define a new data format for the graphical user interface to convert the raw data
+* To configure some alarms on Attribute Value (See RFC-4)
+* To store the Attribute value of a Memorized Attribute (See RFC-4)
+* To configure the labels associated to the possible values of a DevEnum Attribute
+* ...
+
+## Specification
+
+A Property is a strict definition of a pair of key/value:
+  * The Property SHALL have one key, called Property Name
+  * The Property SHALL have one value, which could be empty, called Property Value
+
+There are 4 kinds of Properties:
 
  *  Class Property: a Property related to a Device Class (RFC-2).
  *  Device Property: a Property related to a Device (RFC-2).
  *  Attribute Property: a Property related to an Attribute (RFC-4).
  *  Free Property: a Property defined for the entire Control System (RFC-1) i.e not related to a specific Class, Device or Attribute.
 
-### Use Cases
+A Property MAY be persisted in the Tango Database (See RFC-1 and RFC-6) or into a file if no Tango Database is used.
 
-Here are some Tango Property use cases:
-
-* To connect a real network equipment a Property can define the target IP address or hostname
-* To change the representation of an information in order to be more user friendly a Property can define a new data format for the graphical user interface to convert the raw data
-* To configure some alarms on Attribute Value (See RFC-4)
-* To store the Attribute value of a Memorized Attribute (See RFC-4)
-* To configure the labels associated to a Tango:DevEnum attribute values
-* ...
-
-## Specification
-
-A Tango Property is a strict definition of a pair of key/value:
-  * The Property SHALL have one key, called Property Name
-  * The Property SHALL have one value, which could be empty, called Property Value
-
-A Tango Property MAY be persisted in the Tango Database (See RFC-1 and RFC-6) or into a file if no Tango Database is used.
-
-Tango Device Properties and Tango Class Properties have a Type metadata.
+Device Property and Class Property have a Type metadata.
 The following Types SHALL be supported by Device Properties and Class Properties (See RFC-9):
   * DevBoolean
   * DevShort
@@ -82,7 +82,7 @@ The following Types SHALL be supported by Device Properties and Class Properties
 
 A conforming Tango implementation SHALL provide a way to retrieve a persisted Property Value and to convert it to the above types.
 
-Tango Device Properties and Tango Class Properties MAY have a _Default Value_ which can be used if the Property Value has 
+Device Properties and Class Properties MAY have a _Default Value_ which can be used if the Property Value has 
 not been persisted.
 
 A Tango Device Property MAY be defined as _Mandatory in Database_. 
--- a/5/README.md
+++ b/5/README.md
@@ -67,8 +67,8 @@ A Tango Device Property MAY be defined as _Mandatory in Database_.
 This metadata can be used when the device server programmer requires this Property to be persisted.
 A Device Property defined as _Mandatory in Database_ SHALL not have a Default Value.
 
-Device Property and Class Property have a Type metadata.
-The following Types SHALL be supported by Device Properties and Class Properties (See RFC-9):
+Device Property and Class Property MAY have a Type metadata.
+It is RECOMMENDED to support the following Types for Device Properties and Class Properties (See RFC-9):
   * DevBoolean
   * DevShort
   * DevUShort
@@ -86,7 +86,7 @@ The following Types SHALL be supported by Device Properties and Class Properties
   * DevVarDoubleArray
   * DevVarStringArray
 
-A conforming Tango implementation SHALL provide a way to retrieve a persisted Property Value and to convert it to the above types.
+It is RECOMMENDED to provide a way to retrieve a persisted Property Value and to convert it to the above types.
 
 It is RECOMMENDED to support _Not a Number (NaN)_ as well as _-infinity (-inf)_ and _+infinity (+inf)_ DevFloat and DevDouble values.
 
--- a/5/README.md
+++ b/5/README.md
@@ -1,18 +1,24 @@
 ---
 domain: rfc.tango-controls.org
 shortname: 5/PROPERTY
-name:Property
+name: Property
 status: raw
-editor: Vincent Hardion (vincent.hardion@maxiv.lu.se)
+editor: Gwenaëlle Abeillé <gwenaelle.abeille@synchrotron-soleil.fr>
+contributors:
+  - Reynald Bourtembourg <reynald.bourtembourg@esrf.fr>
+  - Thomas Braun <thomas.braun@byte-physics.de>
+  - Andy Gotz <andy.gotz@esrf.fr>
+  - Vincent Hardion <vincent.hardion@maxiv.lu.se> 
+  - Lorenzo Pivetta <lorenzo.pivetta@elettra.eu>
 ---
 
-This document describes the Property,  a Tango concept representing an element of configuration in order to customise a Tango element. This document describes version 1.0 of the Property.
+This document describes the Tango Property model.
 
-See also: Y/OtherTemplate
+See also: 1/TANGO, 2/Device, 4/Attribute, 6/Database, 9/DataTypes
 
 ## Preamble
 
-Copyright (c) 2019 MAX IV Laboratory.
+Copyright (c) 2019 Tango Controls
 
 This Specification is free software; you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation; either version 3 of the License, or (at your option) any later version. This Specification is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details. You should have received a copy of the GNU General Public License along with this program; if not, see <http://www.gnu.org/licenses>.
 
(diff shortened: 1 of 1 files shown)


bourtemb:
As a follow-up of what I said during the last RFC teleconf meeting, I have rewritten from scratch the RFC/5 about the Properties.
I created a new Pull Request from a new branch from my fork.
If you prefer we rather continue the work in #30 and in the previous branch, please tell me.

5/README.md
```diff
@@ -22,46 +28,76 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
-Additionally, it aims to:
-
-* Provide the initial information when one starts a Tango Device
-
-* Be usable as keeping global variable in the Tango control system
-
-* Provide meta data complementing any data in Tango
+There are 4 kinds of Tango Properties:
 
+ *  Class Property: a Property related to a Device Class (cf RFC-2).
```
lorenzopivetta: For consistency with the following... s/(cf RFC-2)/(RFC-2)/

5/README.md
```diff
@@ -22,46 +28,76 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
+ *  Device Property: a Property related to a Device (RFC-2).
+ *  Attribute Property: a Property related to an Attribute (RFC-4).
+ *  Free Property: a Property defined for the entire Control System (RFC-1) i.e not related to any Class, Device or Attribute.
```
lorenzopivetta: What about replacing "any Class, ..." with " a specific Class, ...", as a Free Property can be used everywhere even if it not related to.

5/README.md
```diff
@@ -22,46 +28,76 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
-
-
-A Tango Property is a strict definition of a pair of key/value
-* The Property SHALL have one key, called Property Name
-* The Property SHALL have one value
+A Tango Property is a strict definition of a pair of key/value:
+  * The Property SHALL have one key, called Property Name
+  * The Property SHALL have one value, which could be empty, called Property Value
+
+A Tango Property MAY be persisted in the Tango Database of the concerned Tango Control System (See RFC-1 and RFC-6) or into a file if no Tango Database is used.
```
lorenzopivetta: Suggest to remove "of the concerned Tango Control System" as implicitly implied...
Moreover, is it really the case a Property MAY be persisted? Being configuration parameters cannot see much value in not saving a Property... nor a Property Value.

5/README.md
```diff
@@ -22,46 +28,76 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
+  * DevDouble
+  * DevString
+  * DevVarShortArray
+  * DevVarLongArray
+  * DevVarLong64Array
+  * DevVarFloatArray
+  * DevVarDoubleArray
+  * DevVarStringArray
+
+A conforming Tango implementation SHALL provide a way to retrieve a persisted Property Value and to convert it to the above types at least.
```
lorenzopivetta: Suggest to remove "at least"

5/README.md
bourtemb: We can imagine having these device properties being passed as argin parameters to the device server for instance.

lorenzopivetta: ...mmmh... of course it's feasible, but is it really desirable for Tango? Not sure about this, but it makes me somehow uncomfortable...

5/README.md
bourtemb: The idea behind this " at least" was to avoid to limit ourselves to these types.
As a device property is a configuration parameter, we can imagine extending this in the future with some other types very useful in the astronomy world for instance, like images.
But I agree with you, we should define this list to ensure the clients and servers can both support all the types. If we add the support for images, it would be useful for generic tools like jive to be able to display these images.
If we want to add new types, we can update the specs in the future...
I'll follow your suggestion.

5/README.md
bourtemb: > ...mmmh... of course it's feasible, but is it really desirable for Tango? Not sure about this, but it makes me somehow uncomfortable...

It could be useful for tests for instance.

bourtemb: > A Tango Property MAY be persisted....

The idea here as well is that the property is not always persisted.
If you define a Default property value and this default is reasonable for your current use case for all of your devices, you don't need to persist the property value.

5/README.md
bourtemb: Done in https://github.com/tango-controls/rfc/pull/59/commits/898755d7f07eb6dfdf5bcb2fe56fbb2de7d394c4

5/README.md
bourtemb: Done in https://github.com/tango-controls/rfc/pull/59/commits/898755d7f07eb6dfdf5bcb2fe56fbb2de7d394c4

5/README.md
bourtemb: Done in https://github.com/tango-controls/rfc/pull/59/commits/898755d7f07eb6dfdf5bcb2fe56fbb2de7d394c4

5/README.md
```diff
@@ -22,46 +28,76 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
+A Property represents a configuration parameter in Tango.
 
-Additionally, it aims to:
-
-* Provide the initial information when one starts a Tango Device
-
-* Be usable as keeping global variable in the Tango control system
-
-* Provide meta data complementing any data in Tango
+There are 4 kinds of Tango Properties:
```
hardion: These definitions are already specification. They can be moved to the specification paragraph.

5/README.md
```diff
@@ -22,46 +28,76 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
-* To connect a real network equipment a Property can define the target IP address
-
-* To change the representation of an information in order to be more user friendly a Property can define a new data format for the graphical user interface to convert the raw data.
+Here are some Tango Property use cases:
 
+* To connect a real network equipment a Property can define the target IP address or hostname
+* To change the representation of an information in order to be more user friendly a Property can define a new data format for the graphical user interface to convert the raw data
+* To configure some alarms on Attribute Value (See RFC-4)
+* To store the Attribute value of a Memorized Attribute (See RFC-4)
+* To configure the labels associated to a Tango:DevEnum attribute values
```
hardion: Maybe you can skip the "Tango:" but use Attribute Value

5/README.md
```diff
@@ -22,46 +28,76 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
-
-
-A Tango Property is a strict definition of a pair of key/value
-* The Property SHALL have one key, called Property Name
-* The Property SHALL have one value
+A Tango Property is a strict definition of a pair of key/value:
+  * The Property SHALL have one key, called Property Name
+  * The Property SHALL have one value, which could be empty, called Property Value
+
+A Tango Property MAY be persisted in the Tango Database (See RFC-1 and RFC-6) or into a file if no Tango Database is used.
```
hardion: Should a property not be persisted in any case?

5/README.md
```diff
@@ -22,46 +28,76 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
+ *  Free Property: a Property defined for the entire Control System (RFC-1) i.e not related to a specific Class, Device or Attribute.
 
 ### Use Cases
 
-There are many use cases for the usage of a Property:
-
-* To connect a real network equipment a Property can define the target IP address
-
-* To change the representation of an information in order to be more user friendly a Property can define a new data format for the graphical user interface to convert the raw data.
+Here are some Tango Property use cases:
```
hardion: you can skip Tango from Tango Property

5/README.md
```diff
@@ -22,46 +28,76 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
-* data are represented in the form of attributes.
-* actions are represented in the form of commands.
-
-Its model can be represented as a defined tree which each elements are from a defined types: Class, Device, Property, Attribute, Command following the rules below:
-
-
-A Tango Property is a strict definition of a pair of key/value
-* The Property SHALL have one key, called Property Name
-* The Property SHALL have one value
+A Tango Property is a strict definition of a pair of key/value:
```
hardion: Tango Property -> Property

5/README.md
hardion: Tango Property -> Property

5/README.md
```diff
@@ -22,46 +28,76 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
+  * The Property SHALL have one key, called Property Name
+  * The Property SHALL have one value, which could be empty, called Property Value
+
+A Tango Property MAY be persisted in the Tango Database (See RFC-1 and RFC-6) or into a file if no Tango Database is used.
+
+Tango Device Properties and Tango Class Properties have a Type metadata.
```
hardion: Device Property and Class Property

5/README.md
```diff
@@ -22,46 +28,76 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
+The following Types SHALL be supported by Device Properties and Class Properties (See RFC-9):
```
hardion: Actually Property Value are only string. 
A Tango client is unable to know the type of Property.
Adding property from Jive does not require a type. db.get_device_property('...','...') does not reply with the expected type.
From I understand it's only implementation specific to convert property into a variable with an expected type.

5/README.md
```diff
@@ -22,46 +28,76 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
+  * DevVarShortArray
+  * DevVarLongArray
+  * DevVarLong64Array
+  * DevVarFloatArray
+  * DevVarDoubleArray
+  * DevVarStringArray
+
+A conforming Tango implementation SHALL provide a way to retrieve a persisted Property Value and to convert it to the above types.
+
+Tango Device Properties and Tango Class Properties MAY have a _Default Value_ which can be used if the Property Value has
```
hardion: Device Properties and Class Properties

5/README.md
bourtemb: As I wrote to Lorenzo, a default property value can be defined in POGO and if this default value is good for a particular device, the property does not need to be persisted in the DB for this specific use case. So there are cases where a property does not need to be persisted... This is why I've written "MAY be persisted".
It is confusing it seems. Any suggestion to improve this?

lorenzopivetta: It is sort of borderline case... So, maybe it would be useful to describe the mechanism a bit in the rfc, I mean the default value in POGO/source code, even if it is not exactly specification...

5/README.md
bourtemb: > Actually Property Value are only string.

Property values are currently persisted as strings, in our **current** implementation of the Database server.
We could imagine having another Database Server implementation storing them as something different than a string.

In cppTango implementation, there is an abstraction class named DbDatum. There is also another type named DbData. which is a vector of DbDatum.
When we try to get a property value from the Database, we invoke some methods which will put the property value inside a DbDatum object at the end.

The DbDatum implementation provides some extraction operators to be able to extract the property value to many different types.

So, in theory, we could persist the properties in any format we want and keep the same interface on the C++ side (I don't know for the other languages).

In C++, the code generated by Pogo to get the properties is using the DbData/DbDatum classes and these extraction operators (operator>>) to convert the DbDatum to the property types defined by Pogo for these properties.

The main issue here, is to clearly define the scope of these specifications.
If we place ourselves from the device server programmer point of view, we should not say that the property values are strings. I think we don't need to put this restriction.
The tricky question is: from which point of view should we define these specifications?

> A Tango client is unable to know the type of Property.

Indeed. This is currently the case. It could probably be useful to have it in the future.
The Device server on the other side, knows the type of the property.

> Adding property from Jive does not require a type. db.get_device_property('...','...') does not reply with the expected type.

True for this part. But then the DbDatum extraction operators are used to extract the DbDatum to the expected property type. The property type is something known by the device server because the property type is something we define with POGO. It is part of the definition of the device.

> From I understand it's only implementation specific to convert property into a variable with an expected type.

Since the property type is something which is part of the device definition (it is something stored in the Pogo xmi file), I thought it would be useful to precise that at least the property types currently supported by Pogo SHALL be supported.

5/README.md
bourtemb: > It is sort of borderline case... So, maybe it would be useful to describe the mechanism a bit in the rfc, I mean the default value in POGO/source code, even if it is not exactly specification...

Well, there is already this sentence which can probably be improved:

> Device Properties and Class Properties MAY have a _Default Value_ which can be used if the Property Value has not been persisted.

lorenzopivetta: Yes. But it is further on in the spec. What about moving the sentence "A Tango Property MAY be persisted in..." after this one? Moreover, if the Property value is not persisted, wouldn't the default value be used... by default? Propose to write the two sentences as:
- A Tango Property MAY be persisted in the Tango Database (See RFC-1 and RFC-6) or into a file if no Tango Database is used.
- Device Properties and Class Properties MAY have a Default Value which _is_ used if the Property Value has not been persisted.
(emphas_is_ just for this comment)

5/README.md
hardion: I am agree that has to change in the future and this part will fit the purpose.
Yet today Property in the Tango protocol has no type (unfortunately ;-) . You can set a property with a string even if the Tango Device implementation expect a float. There is no mechanism to protect in the Tango world unless explicitly in the code of the Device (generated or not).
And if a tango implementation use or not type for property it does not change anything for the rest of the Tango control system.

xmi file is used by pogo to generate developer code and create helper to convert easily to an expected type. It's not like an Attribute where to have to specify the type.

5/README.md
bourtemb: > Yes. But it is further on in the spec. What about moving the sentence "A Tango Property MAY be persisted in..." after this one?

It was in this order because I thought it was better to have mentioned the possibility for a property to be persisted before.

> Moreover, if the Property value is not persisted, wouldn't the default value be used... by default? Propose to write the two sentences as:
> 
>     * A Tango Property MAY be persisted in the Tango Database (See RFC-1 and RFC-6) or into a file if no Tango Database is used.
> 
>     * Device Properties and Class Properties MAY have a Default Value which _is_ used if the Property Value has not been persisted.
>       (emphas_is_ just for this comment)

Your proposal sounds good to me! The 2 sentences will be regrouped together.

5/README.md
bourtemb: > I am agree that has to change in the future and this part will fit the purpose.
> Yet today Property in the Tango protocol has no type (unfortunately ;-) . 

But what is the definition of the Tango protocol? Aren't we describing the concepts here too?
I think the property type is part of the Tango concepts.

> You can set a property with a string even if the Tango Device implementation expect a float. 

True. But you can set it with a double too ;-). Still in the concept, a device property has a type.
This drives the behaviour of the device server. If you change the type of a device property and you don't change the value in the DB, the behaviour might change because the persisted property value might be interpreted differently. So the type is important.

> There is no mechanism to protect in the Tango world unless explicitly in the code of the Device (generated or not).

True. There is no strong restrictions here. Still I would not say that property values are string.

One of the original and main goals of a Device property is to be used in a Device server code to configure it. The type is important for the device server, at least in C++ and Java.

> And if a tango implementation use or not type for property it does not change anything for the rest of the Tango control system.

Still for the device server programmer, you should be able to store properties into variables having certain types. And this is one of the key features in some languages.
I still have difficulties to draw the lines of the scope of these specifications. :-(

> xmi file is used by pogo to generate developer code and create helper to convert easily to an expected type. It's not like an Attribute where to have to specify the type.

Tango was originally designed with Pogo in mind.
There are many Tango concepts and features which are there thanks to Pogo.
It is the same for the _mandatory in database_ properties for instance. It is the generated code from Pogo which is doing the check that the property value is stored in the DB.
But still I think the concept of _mandatory in database_ property is a feature of Tango, even though it does not necessary need to be implemented in a Tango kernel library. The combination of Pogo and kernel library provides this feature.
The xmi file defining a Tango device is something which can be used as a basis to regenerate code for another implementation of Tango.

The difficulty is to define where we draw the line for these specifications:
From a C++ device server programmer point of view, a persisted property value is a DbDatum, which can eventually be converted to a supported property type.
From the cppTango library point of view, the library gets the value of the persisted device properties for instance, via the result of DbGetDeviceProperty command executed on the TangoDatabase server, which is returning a DevVarStringArray.
The library will take care of converting this DevVarStringArray into a DbData.

This means that if we don't want to change the cppTango library, a compatible DatabaseDs device server should provide this DbGetDeviceProperty command which returns a DevVarStringArray.

On the other hand, if we consider only the source code compatibility on the device server side, we could imagine changing the way cppTango, JTango and pyTango retrieve the persisted properties. these libraries could talk to another service for instance, using a command different than DbGetDeviceProperty on the DatabaseDs, to retrieve device properties. It could even talk to different services depending on some parameters and still the interface on the device server side could stay the same. This new implementation should obviously come with some tools to be able to edit and visualize easily the persisted data.

So where do we draw the line?

bourtemb: > > You can set a property with a string even if the Tango Device implementation expect a float.
> 
> True. But you can set it with a double too ;-). Still in the concept, a device property has a type.
> This drives the behaviour of the device server. If you change the type of a device property and you don't change the value in the DB, the behaviour might change because the persisted property value might be interpreted differently. So the type is important.

Correction: the property type _(for device and class properties)_ is important **from a Device server programmer perspective**.

bourtemb: @hardion, it is true that the sentence:
"Device Property and Class Property have a Type metadata." is a bit confusing because you currently have no way to know this type from a client perspective.
The type is not persisted in the DB for instance and the device wizard from jive for instance does not provide this type so I understand your comment and that this sentence is a bit misleading.
What would you suggest?

hardion: @bourtemb 
I recognise that my comment about Tango Protocol was inappropriate.
My suggestion would be to keep the  "type" part in a separate branch v10 and open an issue with the tag "v10". For me this a valid request. Maybe some questions will be raised like how to handle the migration 9 ->10 (type mandatory or not, if yes what default type ).

5/README.md
```diff
@@ -82,13 +88,6 @@ The following Types SHALL be supported by Device Properties and Class Properties
...
 A conforming Tango implementation SHALL provide a way to retrieve a persisted Property Value and to convert it to the above types.
 
-Device Properties and Class Properties MAY have a _Default Value_ which can be used if the Property Value has 
-not been persisted.
-
-A Tango Device Property MAY be defined as _Mandatory in Database_. 
-This metadata can be used when the device server programmer requires this Property to be persisted.
-A Device Property defined as _Mandatory in Database_ SHALL not have a Default Value.
-
 It is RECOMMENDED to support _Not a Number (NaN)_ as well as _-infinity (-inf)_ and _+infinity (+inf)_ DevFloat and DevDouble values.
```
hardion, line 91: I am confuse here since these values are text and not real. Although I understand the end goal it suppose that you set the property as string.
@piogor these values should also be specify in RFC-9 but I guess it will be express like NaN value such as the square root of negative numbers or the result of 0/0.

bourtemb: > I am confuse here since these values are text and not real. 

No, these values are not text, at least this is not what I meant. Here what I wanted to say is that the property mechanism should allow to handle the NaN, -infinity and +infinity values. So this means that a device server programmer should be able to read a property with these special values and to interpret them correctly as NaN, -infinity and +infinity in the device server code (as defined in standards like https://en.wikipedia.org/wiki/IEEE_754).
This also means that the Tango implementation should be able to persist this kind of values.
It can be persisted as  special strings as it is now by the current implementation of the Database server but another Database server implementation could very well decide to persist it as float or double.

> Although I understand the end goal it suppose that you set the property as string.

Could you elaborate on this point? What do you mean with " you set the property as string" .
Who is "you" in this sentence?  Do you mean it should be persisted as a string?

Ingvord:
I really like the way this RFC is currently written. @bourtemb , great job!

bourtemb:
Hi @gwen-soleil, this PR is ready for editor review.

PR #57, opened 2019-11-07
WIP RFC-9 - Data types

piogor:
Implementation of #56.

PR #61, opened 2019-11-22
RFC 9 datatypes descriptive variant

piogor:
This is alternative to #57 proposition for datatypes specification


piogor:
Hi all, especially @hardion, @bourtemb, @mliszcz, @lorenzopivetta, @gwen-soleil, @srgblnch.  I am sending this PR with a beginning of the alternative proposition of a way of specifying data types, asking for your opinion before I follow with more types.

hardion (commented):
I found the new format a bit more explicit.
And maybe the text representation is not necessary anymore.
For the range my preference goes for the notation from (2^15) - 1 to -(2^15)

9/README.md
```diff
@@ -0,0 +1,139 @@
...
+* DevBoolean SHALL be defined as follows:
+
+```ABNF
+DevBoolean = "DEVBOOLEAN" | "DevBoolean"
+```
+
+* Values which DataType is DevBoolean SHALL be mapped to dev-boolean-value specification:
+
+```abnf
+dev-boolen-value = BIT
```
Ingvord: Well, that is very dangerous assumption. In Java, for instance, **boolean** is 1 BYTE, not a BIT, in fact in C++ it is the same, see for instance on [SO](https://stackoverflow.com/questions/4626815/why-is-a-boolean-1-byte-and-not-1-bit-of-size). 

I would rather say "DevBoolean MUST be mappable to downstream language boolean spec"

piogor: Yes, I will correct this.

hardion:
Maybe it’s better to give the range of the bit type :-)
And don’t forget that the ABNF is mainly for text representation and does not assume the underline container. It’s more or less the same case with double, long, … with 32/64 bits, LSB/MSB, ….

/Vincent

On 25 Nov 2019, at 10:51, Ingvord <notifications@github.com<mailto:notifications@github.com>> wrote:


@Ingvord commented on this pull request.

________________________________

In 9/README.md<https://github.com/tango-controls/rfc/pull/61#discussion_r350079597>:

> +### Numeric
+
+#### DevBoolean
+
+DevBoolean is a data type which values represents logical state of true or false.
+
+* DevBoolean SHALL be defined as follows:
+
+```ABNF
+DevBoolean = "DEVBOOLEAN" | "DevBoolean"
+```
+
+* Values which DataType is DevBoolean SHALL be mapped to dev-boolean-value specification:
+
+```abnf
+dev-boolen-value = BIT


Well, that is very dangerous assumption. In Java, for instance, boolean is 1 BYTE, not a BIT, in fact in C++ it is the same, see for instance on SO<https://stackoverflow.com/questions/4626815/why-is-a-boolean-1-byte-and-not-1-bit-of-size>.

I would rather say "DevBoolean MUST be mappable to downstream language boolean spec"

9/README.md
```diff
@@ -0,0 +1,227 @@
...
+ DataType =/ DevVarBooleanArray / DevVarDoubleArray / DevVarFloatArray
+ DataType =/ DevVarShortArray / DevVarLongArray / DevVarLong64Array / DevVarCharArray
+ DataType =/ DevVarStringArray
+ DataType =/ DevVarUShortArray / DevVarULongArray / DevVarULong64Array
+ 
+ DataType =/ DevEncoded / DevVarEncodedArray
+ 
+ DataType =/ DevVarLongStringArray / DevVarDoubleStringArray
+ 
+ DataType =/ DevState
```
bourtemb: DevVarStateArray and DevEnum seem to be missing for the moment.
In tango_const.h from cppTango, I see as well the following types:
- ConstDevString
- DevInt
- DevPipeBlob

For the DevInt type, we need to clarify what it is and whether we really need it.

9/README.md
```diff
@@ -0,0 +1,227 @@
...
+### Numeric
+
+#### DevBoolean
+
+DevBoolean is a data type which values represents logical state of true or false.
+
+* DevBoolean SHALL be defined as follows:
+
+```abnf
+DevBoolean = "DEVBOOLEAN" / "DevBoolean"
```
bourtemb: I don't know whether this is relevant but in cppTango, in enum CmdArgType, the enum associated to DevBoolean type is named DEV_BOOLEAN

9/README.md
piogor: Aren't ConstDevString and DevInt somehow C++ implementation specific? 
These are not defined in the tango.idl. Pogo, allows to use ConstDevSting for command but DevInt is not there at all.

9/README.md
piogor: Thanks, @bourtemb. I will reserach this more, because it applies to other types, too.

9/README.md
bourtemb: > Aren't ConstDevString and DevInt somehow C++ implementation specific?
> These are not defined in the tango.idl.

Good point. So we should probably not mention these types.

9/README.md
piogor: Together with @mliszcz,  we did a research and it seems there are no 'protocol' defined names for types except "CamelCase" strings commonly used in documentation and GUI applications. 

The `DEVBOOLEAN` and `DEV_BOOLEAN` are labels in enums for java and cpp, respectively. Even more, JTango is using also `Tango_DEV_BOOLEAN`. 

It seems, the more important are long values assigned to types, which can be found from both enums and consts defined in [JTango](https://github.com/tango-controls/JTango/blob/14a6527e3ca67339e51a905ef6c55d773ed04cb7/common/src/main/java/fr/esrf/TangoDs/TangoConst.java#L268) and probably related to the index of  [CmdArgTypeName array](https://github.com/tango-controls/cppTango/blob/4d2ad6555fd6bc6e6603406ea83b6ed0ed797eb3/cppapi/server/tango_const.h.in#L1024 ).

So, I suggest to specify data types names as camel case strings only and add info about associated long values. @hardion, @Ingvord, @t-b, I would be glad having your opinion, too :-)

piogor: Finally, I left only "DevCamelCase" names.

9/README.md
piogor: DevPipeBlob is there, now :-)

piogor:
Hi @gwen-soleil, the PR seems to be ready for editor's review :-)

9/README.md
```diff
@@ -0,0 +1,322 @@
...
+ long-string-value = lvalue svalue 
+ double-string-value = dvalue svalue
+                       ; elements order implied by the above rules MAY be ignored
+```
+;where \<lvalue\> has type DevVarLongArray, \<dvalue\> has type DevVarDoubleArray and \<svalue\> has type 
+DevVarStringArray.
+
+#### State
+
+DavState type standardize a way of describing state of Device. Its values are from predefined set of 14 names.
```
gwen-soleil: Typo fix:
```suggestion
DevState type standardize...  
```

9/README.md
```diff
@@ -0,0 +1,322 @@
...
+;where \<lvalue\> has type DevVarLongArray, \<dvalue\> has type DevVarDoubleArray and \<svalue\> has type 
+DevVarStringArray.
+
+#### State
+
+DavState type standardize a way of describing state of Device. Its values are from predefined set of 14 names.   
+
+* DevState name SHALL be according to the following rule:
+```abnf
+ DevState = "DevState
```
gwen-soleil:
```suggestion
 DevState = "DevState"
```

9/README.md
piogor: thanks @gwen-soleil, fix is coming.

9/README.md
piogor: fix is coming

9/README.md
```diff
@@ -0,0 +1,322 @@
...
+```abnf
+ devenum-label = 1*VCHAR
+```  
+
+* DevEnum labels SHALL be unique in the context of an Attribute.
+* Each of DevEnum labels SHALL correspond to one and unique DevShort value. 
+* Related DevShort values SHALL be consecutive and SHALL start with 0. 
+* Values of DevEnum DataType SHALL be transported as values of DevShort.
+
+* For any of Attribute of DevEnum data type, list of Labels SHALL be available as `enum_labels` Attribute property.
```
gwen-soleil:
```suggestion
* For any of Attribute of DevEnum data type, list of Labels SHALL be available as `enum_labels` Attribute property (cf RFC4). 
```

9/README.md
```diff
@@ -0,0 +1,322 @@
...
+ DevFailed = "DevFailed"
+```
+
+* Any value of DevFiled type SHALL follow \<dev-failed-value\> rule:
+```abnf
+ dev-failed-value = 1*error 
+
+ error = reason severity desc origin
+
+ reason = *VCHAR ; SHALL be the reason of the failur
```
gwen-soleil:
```suggestion
 reason = *VCHAR ; SHALL be the reason of the failure
```

9/README.md
```diff
@@ -0,0 +1,322 @@
...
+ severity = "WARN" | "ERR" | "PANIC"
+ desc = *VCHAR ; SHOULD describe the failur in more details, eg. call stack
```
gwen-soleil:
```suggestion
 desc = *VCHAR ; SHOULD describe the failure in more details
```
For me:
-  the call stack is in the "origin" field and not the "desc" field.
- desc is a message describing what is occuring in english.
- "reason" is string without spaces that is standardized for some errors thrown by the API. examples: "API_AttrOptProp", "API_AttrNotPolled","API_CmdNotPolled", "API_CommandNotFound", .... 
@bourtemb What do you think? Should we list all reasons of the API in the Tango RFC? I have found 39 of them in the Java code.

9/README.md
```diff
@@ -0,0 +1,322 @@
...
+a Tango Controls systems irrespectively to its encoding in a transport protocol or computers' CPU architecture.
+It also allows to distinguish different kind of numeric data, string data, enumerations or multi-values data. 
+
+### Use Cases
+
+There are many use cases for Data Types. Some of them are listed below:
+
+* To specify expected data format for attributes, commands and pipes.
+
+* To display in the same way a numeric values read from a devices when some of them are using big-endian encoding and
```
gwen-soleil:
```suggestion
* To display in the same way a numeric values read from a device when some of them are using big-endian encoding and 
```

9/README.md
bourtemb: @gwen-soleil, I think it's good to standardize the reason field for the most common errors, and for this I think it's good to have a place where these errors are listed, but I don't think this RFC is the best location to list them all. Maybe we can add a comment as you did saying that the reason field is standardized for some errors thrown by the API, with some examples.
We would probably need another RFC (_Errors_ or something like that) to list all of these standard errors.

9/README.md
```diff
@@ -0,0 +1,322 @@
...
+
+* Any value of DevFiled type SHALL follow \<dev-failed-value\> rule:
+```abnf
+ dev-failed-value = 1*error 
+
+ error = reason severity desc origin
+
+ reason = *VCHAR ; SHALL be the reason of the failure
+ severity = "WARN" | "ERR" | "PANIC"
+ desc = *VCHAR ; SHOULD describe the failur in more details, eg. call stack
```
bourtemb: The call stack is actually usually deduced from the origin fields of the **list** of error**s**.
For me, `origin `contains only 1 method name (The one from which the exception is thrown), this is not a call stack. But a call stack can be deducted from the origin fields of the list of errors.

9/README.md
piogor: I have checked the tango_const.h and there is about 130 definitions of API_... reasons. However, there is no mechanism for forcing usage of only these names for reasons.  Are these compatible between java and c++ implementations?

piogor: I suggest changing SHALL to SHOULD, and skip listing all the possible reasons because if there will be a need to add a new one in the RFC will be invalidated and possibly not compatible between java and c++.

9/README.md
```diff
@@ -0,0 +1,322 @@
...
+ DevFailed = "DevFailed"
+```
+
+* Any value of DevFiled type SHALL follow \<dev-failed-value\> rule:
+```abnf
+ dev-failed-value = 1*error 
+
+ error = reason severity desc origin
+
+ reason = *VCHAR ; SHALL be the reason of the failure
```
piogor:
```suggestion
 reason = *VCHAR ; SHOULD be a symbolic name, without whitespace, which points the reason for the failure and is standardized for some errors thrown by the API. Examples: "API_AttrOptProp", "API_AttrNotPolled","API_CmdNotPolled", "API_CommandNotFound"
```

9/README.md
```diff
@@ -0,0 +1,322 @@
...
+ severity = "WARN" | "ERR" | "PANIC"
+ desc = *VCHAR ; SHOULD describe the failur in more details, eg. call stack
+ origin = *VCHAR ; SHALL identify the operation or the tango object which caused the failure
```
piogor:
```suggestion
 origin = *VCHAR ; SHALL identify the operation or the tango object which caused the failure, eg. function name
```

9/README.md
bourtemb: > I have checked the tango_const.h and there is about 130 definitions of API_... reasons. However, there is no mechanism for forcing usage of only these names for reasons. Are these compatible between java and c++ implementations?

I guess many of them are compatible because the Java API was developed after the C++ one and was mainly based on it at the beginning so I guess many errors are common in these 2 languages but I would not be surprised to see some differences.
This is why I find it interesting to have another RFC listing at least the standardized reason fields for the most common errors coming from a Tango implementation, with a description of these errors if it makes sense.
Then, 2 different Tango implementations could refer to this RFC and use consistent reason fields for similar errors.

9/README.md
piogor: @bourtemb and @gwen, what do you think about this?

9/README.md
```diff
@@ -0,0 +1,322 @@
...
+ DevFailed = "DevFailed"
+```
+
+* Any value of DevFiled type SHALL follow \<dev-failed-value\> rule:
+```abnf
+ dev-failed-value = 1*error 
+
+ error = reason severity desc origin
+
+ reason = *VCHAR ; SHOULD be a symbolic name, without whitespace, which points the reason for the failure and is standardized for some errors thrown by the API. Examples: "API_AttrOptProp", "API_AttrNotPolled","API_CmdNotPolled", "API_CommandNotFound"
```
piogor:
```suggestion
 reason = *VCHAR ; SHOULD be a symbolic name, without whitespace, which points the reason for the failure and is standardized 
                        ; for some errors thrown by the API. Examples: "API_AttrOptProp", "API_AttrNotPolled","API_CmdNotPolled", "API_CommandNotFound"
```

9/README.md
piogor: You have a point. I will add this RFC to the list of ones which are waiting to be added :)

9/README.md
```diff
@@ -0,0 +1,322 @@
...
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", 
+and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119).
+
+## Tango Data Types Specification
+
+Data Types are specifying format and limits of pieces of information exchanged between Tango Controls components.   
+
+### Goals
+
+Data Types aims to standardise a way how different kind of information is entered, displayed and processed within
```
bourtemb: Data Types aims => Data Types specification aims or Data Types aim

9/README.md
```diff
@@ -0,0 +1,322 @@
...
+a Tango Controls systems irrespectively to its encoding in a transport protocol or computers' CPU architecture.
```
bourtemb: a Tango Controls systems => a Tango Controls system

9/README.md
```diff
@@ -0,0 +1,322 @@
...
+It also allows to distinguish different kind of numeric data, string data, enumerations or multi-values data.
```
bourtemb: different kind => different kinds?

9/README.md
```diff
@@ -0,0 +1,322 @@
...
+a Tango Controls systems irrespectively to its encoding in a transport protocol or computers' CPU architecture.
+It also allows to distinguish different kind of numeric data, string data, enumerations or multi-values data. 
+
+### Use Cases
+
+There are many use cases for Data Types. Some of them are listed below:
+
+* To specify expected data format for attributes, commands and pipes.
+
+* To display in the same way a numeric values read from a device when some of them are using big-endian encoding and
```
bourtemb: a numeric values => numeric values
                   or    => a numeric value

9/README.md
```diff
@@ -0,0 +1,322 @@
...
+It also allows to distinguish different kind of numeric data, string data, enumerations or multi-values data. 
+
+### Use Cases
+
+There are many use cases for Data Types. Some of them are listed below:
+
+* To specify expected data format for attributes, commands and pipes.
+
+* To display in the same way a numeric values read from a device when some of them are using big-endian encoding and 
+  other use little-endian encoding.
```
bourtemb: other => others

9/README.md
```diff
@@ -0,0 +1,322 @@
...
+
+```abnf
+DevBoolean = "DevBoolean"
+```
+
+* Values which DataType is DevBoolean SHALL carry one bit of information.
+
+#### Integer data types
+
+DevShort, DevLong, DevLong64 represent singed integer values. DevUChar, DevUShort, DevULong and DevULong64 represent
```
bourtemb: singed => signed

9/README.md
```diff
@@ -0,0 +1,322 @@
...
+ DevShort = "DevShort"
+ DevLong = "DevLong"
+ DevLong64 = "DevLong64"
+ DevUChar = "DevUChar"
+ DevUShort = "DevUShort"
+ DevULong = "DevULong"
+ DevULong64 = "DevULong64"
+```
+
+* Values of integer data types SHALL be in defined ranges and the information it cary SHALL have number of bits
```
bourtemb: it cary => they carry?

9/README.md
```diff
@@ -0,0 +1,322 @@
...
+### Numeric
+
+#### DevBoolean
+
+DevBoolean is a data type which values represents logical state of true or false.
+
+* DevBoolean SHALL be defined as follows:
+
+```abnf
+DevBoolean = "DevBoolean"
```
bourtemb: In this specific case, it is not obvious that this line defines the name.

9/README.md
```diff
@@ -0,0 +1,322 @@
...
+ DevLong = "DevLong"
+ DevLong64 = "DevLong64"
+ DevUChar = "DevUChar"
+ DevUShort = "DevUShort"
+ DevULong = "DevULong"
+ DevULong64 = "DevULong64"
+```
+
+* Values of integer data types SHALL be in defined ranges and the information it cary SHALL have number of bits
+  specified in the table below:
```
bourtemb: specified => as specified?

9/README.md
```diff
@@ -0,0 +1,322 @@
...
+ long-string-value = lvalue svalue 
+ double-string-value = dvalue svalue
+                       ; elements order implied by the above rules MAY be ignored
+```
+;where \<lvalue\> has type DevVarLongArray, \<dvalue\> has type DevVarDoubleArray and \<svalue\> has type 
+DevVarStringArray.
+
+#### State
+
+DevState type standardize a way of describing state of Device. Its values are from predefined set of 14 names.
```
bourtemb: standardize => standardizes

9/README.md
```diff
@@ -0,0 +1,322 @@
...
+
+* DevState name SHALL be according to the following rule:
+```abnf
+ DevState = "DevState"
+```
+
+* Value of DevState type SHALL follow the rule \<dev-state-value\>:
+```abnf
+ dev-state-value = "ALARM" / "INSERT" / "STANDBY" / "CLOSE" / "MOVING" / "UNKNOWN" / "DISABLE" / "OFF"
+ dev-state-value =/ "EXTRACT" / "ON" / "FAULT" / "OPEN" / "INIT" / "RUNNING"
```
bourtemb: This is describing only the names of the different states, not the values...
I find this a bit confusing.
In the IDL, DevState is declared as an enum.

9/README.md
```diff
@@ -0,0 +1,322 @@
...
+```abnf
+ devenum-label = 1*VCHAR
+```  
+
+* DevEnum labels SHALL be unique in the context of an Attribute.
+* Each of DevEnum labels SHALL correspond to one and unique DevShort value. 
+* Related DevShort values SHALL be consecutive and SHALL start with 0. 
+* Values of DevEnum DataType SHALL be transported as values of DevShort.
+
+* * For any of Attribute of DevEnum data type, list of Labels SHALL be available as `enum_labels` Attribute property (see [4/Attribute](/4)).
```
bourtemb: 1 star too much here at the beginning of the line?
For any of Attribute of => For any Attribute of

9/README.md
```diff
@@ -0,0 +1,322 @@
...
+```abnf
+ pipe-blob-value = name blob-data
+                   ; elements order implied by the above rule MAY be ignored
+
+ name = 1*VCHAR 
+
+ blob-data = *blob-data-element
+ 
+ blob-data-element = name value inner-blob inner-blob-name 
+                     ; elements order implied by the above rule MAY be ignored
```
bourtemb: inner-blob-name is not described it seems

9/README.md
```diff
@@ -0,0 +1,322 @@
...
+ 
+ blob-data-element = name value inner-blob inner-blob-name 
+                     ; elements order implied by the above rule MAY be ignored    
+
+ inner-blob = blob-data  
+```
+
+#### Exceptions 
+
+Results of failed operations (exceptions) within the Tango Controls are send as values/messages of DevFailed data type.
```
bourtemb: are send => are sent

bourtemb (commented):
Please have a look at my comments (many typo fixes proposals but not only)

9/README.md
piogor: Solved by this [commit](https://github.com/tango-controls/rfc/pull/61/commits/cf2b7f87ab36e3ce680c36eba2c20120441dca4a)

9/README.md
piogor: Solved with [this commit](https://github.com/tango-controls/rfc/pull/61/commits/2528659c31865cbbd06df59f5ce867dc34efce0a)

9/README.md
piogor: Hi @bourtemb, [this commit](https://github.com/tango-controls/rfc/pull/61/commits/c7db89be1dbaefbb920c464323883db1df6175e9) proposes the change. Please check if it is ok.

9/README.md
```diff
@@ -0,0 +1,325 @@
...
+ DevFloat = "DevFloat"
+ DevDouble = "DevDouble"
+```
+
+* Values of DevFloat type SHALL be of IEEE 754 single precision floating-point format and MAY be encoded in 32 bits.
+* Values of DevDouble type SHALL be of IEEE 754 double precision floating-point format and MAY be encoded in 64 bits.
+
+#### String data type
+
+DevString values represents strings (sequence of characters).
```
bourtemb: represents => represent

9/README.md
```diff
@@ -0,0 +1,325 @@
...
+#### Exceptions 
+
+Results of failed operations (exceptions) within the Tango Controls are sent as values/messages of DevFailed data type.
+
+* DevFailed type name is as follows:
+```abnf
+ DevFailed = "DevFailed"
+```
+
+* Any value of DevFiled type SHALL follow \<dev-failed-value\> rule:
```
bourtemb: DevFiled => DevFailed

9/README.md
```diff
@@ -0,0 +1,325 @@
...
+
+```abnf
+ pipe-blob-value = name blob-data
+                   ; elements order implied by the above rule MAY be ignored
+
+ name = 1*VCHAR 
+
+ blob-data = *blob-data-element
+ 
+ blob-data-element = name value inner-blob inner-blob-name
```
bourtemb: I think it is not clear what `value ` is here.  (Scalar or array of any basic Tango type)
I also think inner-blob-name does not make sense when the inner-blob is empty.

bourtemb (commented):
2 more typos & 1 comment

9/README.md
piogor: solved with recent commit

9/README.md
piogor: @bourtemb, this [commit](https://github.com/tango-controls/rfc/pull/61/commits/5fb3d9dce38176bc6b4339f208959f8f8b7f6df3) adds the missing `value` specification.

piogor: Regarding the `inner-blob-name`, even it has no meaning in case of inner-blob is empty, the IDL specifies it as a part of the [DevPipeDataElt structure](https://github.com/tango-controls/tango-idl/blob/1e5edb84d966814ad367f2674ac9a5658b6724ac/tango.idl#L555).

piogor:
@gwen-soleil and @bourtemb, I think it is now ready :-). I am removing WIP status.

9/README.md
```diff
@@ -303,6 +303,10 @@ DevPipeBlob is a Data Type to transfer data related to Pipes (see 7/Pipe).
  inner-blob-name = name
 ```
 
+Where \<value\> SHALL be a value of one of the following data types: DevVarBooleanArray, DevVarShortArray, DevVarLongArray,			
+DevVarLong64Array, DevVarFloatArray, DevVarDoubleArray, DevVarCharArray, DevVarUShortArray,	DevVarULongArray,		
+DevVarULong64Array, DevVarStringArray, DevVarStateArray, DevState, DevVarEncodedArray, DevBoolean.
```
bourtemb: What about scalar DevDouble, DevFloat, ...?

piogor: Apparently, according to IDL, there are no scalar data types. It is [AttrValUnion](https://github.com/tango-controls/tango-idl/blob/1e5edb84d966814ad367f2674ac9a5658b6724ac/tango.idl#L363).

bourtemb: According to the IDL. Yes, you're right.
I've seen the doc (https://tango-controls.readthedocs.io/en/latest/development/client-api/cpp-client-programmers-guide.html#reading-writing-device-pipe) is mentioning scalar data types too.
I think scalar are simply transported in arrays.

There is something confusing, I think.
It is the presence of DevBoolean type.
It is there only for the ATT_NO_DATA case.

piogor: Yes, DevBoolean is only to poit there are no data :-)

bourtemb: [DevicePipeBlob::get_data_elt_type() method in cppTango](https://github.com/tango-controls/cppTango/blob/tango-9-lts/cppapi/client/devapi_pipe.cpp#L508-L517) distinguishes scalar and spectrum types.
So, from the user point of view, it seems like the data elements can be scalar Tango data types too.

bourtemb: And when looking deeper into cppTango code, it seems like the inner_blob_name is set to "scalar" string when the element is a scalar.

https://github.com/tango-controls/cppTango/blob/tango-9-lts/cppapi/client/devapi_pipe.cpp#L507-L513

and

https://github.com/tango-controls/cppTango/blob/tango-9-lts/cppapi/server/tango_const.h.in#L106-L107

piogor: Yes, you are right. I think what matters here is users' perspectives, not a corba implementation. I will add the missing types.

bourtemb: It is interesting to note that the corresponding [JTango PipeDataElement::getType](https://github.com/tango-controls/JTango/blob/3c826dc478e1a4c2f3c6f2d154393184801f3e29/common/src/main/java/fr/esrf/TangoApi/PipeDataElement.java#L441) method does not seem to distinguish scalar and array types.

The PipeDataElement constructors seem to construct some DevPipeDataElt objects which are using a parameter which can be SCALAR or ARRAY, so this seems to be important at some point in Java too.

piogor: I have added the [scalars](https://github.com/tango-controls/rfc/pull/61/commits/d804c9f03fa5a062695a9b359b763a0fc8884ce9)

piogor: Should we create a list of potential pitfalls discovered during writing RFCs?

bourtemb: Why not? Where do you want to store them?

9/README.md
```diff
@@ -303,9 +303,12 @@ DevPipeBlob is a Data Type to transfer data related to Pipes (see 7/Pipe).
  inner-blob-name = name
 ```
 
-Where \<value\> SHALL be a value of one of the following data types: DevVarBooleanArray, DevVarShortArray, DevVarLongArray,			
-DevVarLong64Array, DevVarFloatArray, DevVarDoubleArray, DevVarCharArray, DevVarUShortArray,	DevVarULongArray,		
-DevVarULong64Array, DevVarStringArray, DevVarStateArray, DevState, DevVarEncodedArray, DevBoolean.
+Where \<value\> SHALL be a value of one of the following data types: 
+DevVoid, DevBoolean, DevShort, DevLong, DevLong64, DevFloat, DevDouble, DevChar, DevUShort,	DevULong,
```
bourtemb: I think in the case of pipe, DevVoid type does not make sense. I think it should be removed.

9/README.md
```diff
@@ -0,0 +1,332 @@
...
+ name = 1*VCHAR 
+
+ blob-data = *blob-data-element
+ 
+ blob-data-element = name value inner-blob inner-blob-name 
+                     ; elements order implied by the above rule MAY be ignored    
+
+ inner-blob = blob-data  
+
+ inner-blob-name = name
```
bourtemb: I propose to remove this inner-blob and inner-blob-name to have something like:

`blob-data-element = name value / name blob-data`

bourtemb:
9/README.md should be renamed in 9/Datatypes.md or something similar

9/README.md
piogor: I don know yet

9/README.md
piogor: I suggest then `blob-data-element = name value / blob-data`

piogor:
> 
> 
> 9/README.md should be renamed in 9/Datatypes.md or something similar

Yes, you are right :)

9/DataTypes.md
```diff
@@ -295,16 +295,12 @@ DevPipeBlob is a Data Type to transfer data related to Pipes (see 7/Pipe).
 
  blob-data = *blob-data-element
  
- blob-data-element = name value inner-blob inner-blob-name 
-                     ; elements order implied by the above rule MAY be ignored    
-
- inner-blob = blob-data  
-
- inner-blob-name = name
+ blob-data-element = name value / blob-data
```
bourtemb: When looking at the [doc](https://tango-controls.readthedocs.io/en/latest/development/client-api/cpp-client-programmers-guide.html#reading-writing-device-pipe), it seems like a name is missing in your new line.
This is why I suggested:

`blob-data-element = name value / name blob-data `

But maybe the doc is wrong?
I'm still quite a newbie with Tango pipes.

bourtemb: or maybe this would be also correct in abnf?:

`blob-data-element = name (value / blob-data) `

I think what you've written is equivalent to:
` blob-data-element = (name value) / blob-data `

piogor:
```suggestion
 blob-data-element = name (value / blob-data) 
```

PR #62, opened 2019-11-24
The pipe model (RFC-7)

HelgaMond:
README.md is added


7/README.md
```diff
@@ -0,0 +1,95 @@
...
+
+This Specification is free software; you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation; either version 3 of the License, or (at your option) any later version. This Specification is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details. You should have received a copy of the GNU General Public License along with this program; if not, see <http://www.gnu.org/licenses>.
+
+This Specification is a [free and open standard](http://www.digistan.org/open-standard:definition) and is governed by the Digital Standards
+Organization's [Consensus-Oriented Specification System](http://www.digistan.org/spec:1/COSS).
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119).
+
+
+# Tango Pipe specificatiom
```
bourtemb:
```suggestion
# Tango Pipe specification
```

7/README.md
```diff
@@ -0,0 +1,95 @@
...
+
+This specification describes the Tango Pipe feature of a Device. 
+
+## Goals
+
+Tango Pipe is an Attribute [RFC-4] with custom data structure.
```
bourtemb: I would not say that a Pipe **is** an Attribute, I think this is misleading.
This section of the documentation might help:
https://tango-controls.readthedocs.io/en/latest/development/device-api/device-server-model.html#the-tango-pipes

> In contrary of commands or attributes, a TANGO pipe does not have a pre-defined data type. Data transferred through pipes may be of any basic Tango data type (or array of) and this may change every time a pipe is read or written.

7/README.md
```diff
@@ -0,0 +1,95 @@
...
+Pipe with writeType="PIPE_READ" MUST be read only.
+
+Pipe data structure MUST be:
+
+ * Key(s) -> Value(s)
+
+```abnf
+Key = DevString [link to RFC-9 Data Types]
+
+Value = An Arary of Any Tango compliant type or Pipe [link to RFC-9 Data Types]
```
bourtemb: I would refer to this section of the documentation to help to write the abnf:
https://tango-controls.readthedocs.io/en/latest/development/client-api/cpp-client-programmers-guide.html#reading-writing-device-pipe

I think we should talk about the notion of DataElement and PipeBlob

bourtemb (commented):
Thanks @HelgaMond and @Ingvord for tackling this RFC related to something which is still mysterious for most of the Tango users (including me).

7/README.md
HelgaMond: @bourtemb Thank you for the comments. 

I've made changes. And I have added an image with "Key-Value" based on the Figure 4.1 https://tango-controls.readthedocs.io/en/latest/development/client-api/cpp-client-programmers-guide.html#reading-writing-device-pipe

The reason is that I don't know whether it is a good idea to use such special words as  "DataElement" or "PipeBlob" in RFC... As I understand, in fact it is something like Map so why not to use common words.

bourtemb: Thanks @HelgaMond for the update.
I think there is an important information missing in the new image you uploaded.
On the right, in the figure from the doc, the data element value can be data or a DevicePipeBlob.
In the new image, data is absent  (This is actually described in the "_abnf_" definition of Value you proposed). So this makes the structure appearing to be infinitely recursive. I find it a bit confusing.
The key-value analogy is not stupid I think but I would not say that a pipe is completely like a map. It is more a special structure. I personally don't like the term "Description" at the higher level. I find it confusing too.
In my opinon, the terms DevicePipeBlob, DataElement are already mentioned in the current documentation and in the Tango IDL (https://github.com/tango-controls/tango-idl/blob/tango-9-lts/tango.idl#L547-L569) so I would simply use these terms.

7/README.md
```diff
@@ -73,6 +74,9 @@ Key = DevString [link to RFC-9 Data Types]
 Value = An Arary of Any Tango compliant type or Pipe [link to RFC-9 Data Types]
```
bourtemb: Value = An Arary ... => Value = An Array

7/README.md
bourtemb: @HelgaMond, I noticed @piogor described the pipes in RFC-9 (Datatypes) too (https://github.com/tango-controls/rfc/pull/61/files), so if you change the names, it would be good to coordinate with @piogor to get some consistent names across the different RFCs.

HelgaMond: @bourtemb Could you have a look at the RFC-7, please? I have made minor changes and as agreed have returned to the "DevicePipeBlob" definition. Thank you!

7/README.md
bourtemb: I recommend to change the following part too:

> Tango Pipe is an Attribute [RFC-4] with custom data structure.

I suggest to replace it with something like:

> Tango Pipe is like an Attribute [RFC-4] with custom data structure.

or

> Tango Pipe is similar to an Attribute [RFC-4] but with custom data structure.

7/README.md
bourtemb: According to the doc, it seems like value can be a scalar of any tango compliant type too (https://tango-controls.readthedocs.io/en/latest/development/client-api/cpp-client-programmers-guide.html#reading-writing-device-pipe)

> In a DevicePipeBlob instance, you have:

    The blob name
    One array of DataElement. Each instance of this DataElement class has:
        A name
        A value which can be either
            Scalar or array of any basic Tango type
            Another DevicePipeBlob

7/README.md
```diff
@@ -0,0 +1,101 @@
+---
+domain: rfc.tango-controls.org 
+shortname: 7/PIPE 
+name: The pipe 
+status: raw 
+editor:
```
bourtemb:
```suggestion
editor:  Reynald Bourtembourg `<at esrf.fr - reynald.bourtembourg>`
```

7/README.md
```diff
@@ -0,0 +1,101 @@
...
+
+Value = An Array of Any Tango compliant type or Pipe [link to RFC-9 Data Types]
+```
+
+In Tango 9 "DevPipeBlob" is used to describe Value (see image below).
+
+![](pipe_Tango_9.png)
+
+
+> **V10 NOTE:** Attribute being a corner case of a Pipe MUST be implemented as a derivative of a Pipe.
```
bourtemb: As in RFC-15 Tango V10 note, I would not be as radical and avoid the usage of the MUST verb here.
I am not even sure that this would be such a good idea to implement attributes as derivative of pipes... This is a possibility that should be studied, but I would not put such radical affirmations here.
How would you do if you would like to implement Tango with a programming language which is not object oriented for instance?

bourtemb (commented):
7/README.md should be renamed 7/Pipe.md

I added a suggestion to add me as an editor and 1 comment about the V10 note

7/README.md
HelgaMond: Thank you for the comment. I have change the sentence. 

But thinking about the languages. As far as I know any language has some structure. What do you mean by asking this question? Could you give an example, please?

7/README.md
HelgaMond: Done! Thank you!

7/README.md
HelgaMond: Oh! Thank you. My mistake. In Java it is array :)

7/README.md
bourtemb: > Thank you for the comment. I have change the sentence.

Thank you

> But thinking about the languages. As far as I know any language has some structure. What do you mean by asking this question? Could you give an example, please?

Sorry, maybe I misinterpreted the term derivative in the sentence. I understood you were suggesting to have an attribute inheriting from pipe in terms of object oriented programming...

7/Pipe.md
```diff
@@ -0,0 +1,101 @@
...
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119).
+
+
+# Tango Pipe specification
+
+This specification describes the Tango Pipe feature of a Device. 
+
+## Goals
+
+Tango Pipe is an Attribute [RFC-4] with custom data structure - the pipe name and pipe description.
```
bourtemb, line 28: I guess we should write pipe with a first upper case letter everywhere => Pipe

bourtemb, line 28: Should we start to add links to the other RFCs?
@piogor what is the convention for the links to the other RFCs?
When looking at https://github.com/tango-controls/rfc/blob/draft/4/Attribute.md, it seems like the convention used there was to create links with the name of the RFC. Here for RFC 4 it would be:
RFC 4/Attribute which would point to https://github.com/tango-controls/rfc/blob/draft/4/Attribute.md I guess

In RFC/4, the links are pointing to the root directory of the RFC, I guess this should be changed to point to the .md file since it is no longer named README.md

bourtemb (approved):
Thanks @HelgaMond  
The content looks good to me.
There is still an outstanding work to be done to add links to other RFCs but this can be done later on when a decision will have been taken in https://github.com/tango-controls/rfc/issues/88.
The question of when using upper case first letters can also be addressed later on when a global harmonization work on all the RFCs will be done.
So, if no objection, I will merge this PR this afternoon after the RFC meeting.

7/Pipe.md
HelgaMond: Dear @bourtemb Thank you very much for your review. For sure the links and using upper case first letters will be updated later. Please, merge the RFC.

PR #81, opened 2020-02-04
WIP RFC-10 Request Reply

hardion:
This PR follows up all development of the Request Reply specification


piogor (approved):
@hardion, I will merge to rebase.

Issue #83, opened 2020-02-10
Second iteration on RFC-14/Logging system


srgblnch:
Once completed the [Pull Request](https://github.com/tango-controls/rfc/pull/52) from the RFC-14, some conversation happen in the periodical meeting.

1) This RFC has the description of what we have, as an specification, but then opens 3 sections (aka C++API, Python API, Java API) that show divergences that the 3 current implementations show. It could be nice to discuss how we can include this information in this specification not loosing that it is an specification (and don't go to deep in the implementation).

2) This RFC also describes a GUI to view logs, that perhaps is better to translate to an specification of how a log viewer could work based on what we have implemented.

3) I would like also to suggest also to change the acronym TLS of Tango Logging System to something that doesn't collide with Transport Layer Security.

PR #84, opened 2020-02-21
Fixing RFC-14 status and header

piogor:


piogor:
@srgblnch, please review this fix and merge, if it is ok.

piogor:
@srgblnch and @hardion, since there is #83, I am closing this PR without merging as irrelevant.

PR #96, opened 2020-05-22
WIP: RFC-10 - Connections

piogor:


10/RequestReply.md
```diff
@@ -103,14 +103,40 @@ NOTE All GROUPED REQUESTS should be logged once with the same information as the
 
 ### Connection management
 
-* The Reqest-Reply protocol SHALL manage client connection.
+The Reqest-Reply protocol SHALL manage client connection to devcie server.
```
piogor:
```suggestion
The Request-Reply protocol SHALL manage client connections to Device Servers.

File: 10/RequestReply.md [hunk 2]
```diff
@@ -103,14 +103,40 @@ NOTE All GROUPED REQUESTS should be logged once with the same information as the
...
+  network endpoints like operating system inter-process communication mechanism.
+* A Client MAY get information about Device Server's endpoints:
+  * From the Tango Database service, if the service is available,
+  * From command-line parameters,
+  * Other way,
+* The endpoint of Database Devcie Server SHOULD be provided to the Client via the `TANGO_HOST` environment variable defined as follows:
+... abnf
+ tango_host = network_address ":" tcp_port 
+...
+
```
piogor:
```suggestion
* To send any REQUESTS to a Device, a Client MUST connect to the Device Server manging the Device.  
* A Device Server SHALL wait for client connections on one or multiple network endpoints.
* The network endpoint SHALL define \<network_address\> and \<tcp_port\>, where:

... abnf
 network_address = ip_address / host_name
 tcp_port = num-val
...

* A Client MAY connect any Device Server on one of Device Server's endpoints.
* If a Client is running on the same host as a Device Server, the client MAY use other than
  network endpoints like operating system inter-process communication mechanism.
* A Client MAY get information about Device Server's endpoints:
  * From the Tango Database service, if the service is available,
  * From command-line parameters,
  * In another suitable way,
* The endpoint of Database Device Server SHOULD be provided to the Client via the `TANGO_HOST` environment variable defined as follows:
... abnf
 tango_host = network_address ":" tcp_port 
...

```

10/RequestReply.md
```diff
@@ -103,15 +103,40 @@ NOTE All GROUPED REQUESTS should be logged once with the same information as the
...
+* A Device Server SHALL wait for client connections on one or multiple network endpoints.
+* The network endpoint SHALL define \<network_address\> and \<tcp_port\>, where:
+
+... abnf
+ network_address = ip_address / host_name
+ tcp_port = num-val
+...
+
+* A Client MAY connect any Device Server on one of Device Server's endpoints.
+* If a Client is running on the same host as a Device Server, the client MAY use other than
```
hardion, line 120: This is specific to the CORBA protocol so I would say the way to connect is delegated to the underline protocol. It means that this RFC should say explicitly what is an underline protocol.

10/RequestReply.md
```diff
@@ -103,15 +103,40 @@ NOTE All GROUPED REQUESTS should be logged once with the same information as the
...
+  network endpoints like operating system inter-process communication mechanism.
+* A Client MAY get information about Device Server's endpoints:
+  * From the Tango Database service, if the service is available,
```
hardion, line 123: I would rather say that the Tango Database specify the way to get the connection information. Other way have to be specify by the Tango Database RFC.

10/RequestReply.md
```diff
@@ -103,15 +103,40 @@ NOTE All GROUPED REQUESTS should be logged once with the same information as the
...
+  * From command-line parameters,
+  * In another suitable way,
```
hardion, line 125: This is not specific enough and can lead to confusion.

10/RequestReply.md
```diff
@@ -103,15 +103,40 @@ NOTE All GROUPED REQUESTS should be logged once with the same information as the
...
+* The network endpoint SHALL define \<network_address\> and \<tcp_port\>, where:
+
+... abnf
+ network_address = ip_address / host_name
+ tcp_port = num-val
+...
+
+* A Client MAY connect any Device Server on one of Device Server's endpoints.
+* If a Client is running on the same host as a Device Server, the client MAY use other than
+  network endpoints like operating system inter-process communication mechanism.
```
piogor, line 121:
```suggestion
```

10/RequestReply.md
piogor: So, I propose to remove this point.

10/RequestReply.md
piogor: I look on [the RFC-6](https://github.com/tango-controls/rfc/blob/draft/6/Database.md#exported-devices) and I thinkit is worth to add that the Device Server SHALL call DbExportDevice to publish the endpoint and DbUnExportDevice to unpublish.

hardion: Good point! In this case we should emphasis the role of the Device Server to use the  Request-Reply to register the new connection.

PR #102, opened 2020-07-31
rfc-10, reconnection policy

piogor:


10/RequestReply.md
```diff
@@ -62,7 +62,16 @@ There are few main use cases for The Request-Reply protocol:
 
 ### Client duty
 #### Reconnection
-TODO Describe here how to check the connection and in which condition should the client reconnect
+
+The client (API?) SHALL manage connections to any Device Servers it is using.
```
piogor: I am not sure whether to use `the client` or `the client API`. @bourtemb, @hardion, @lorenzopivetta, what do you think?

piogor: Aftter discussion on ther meeting, `client` is ok.

PR #109, opened 2020-08-07
WIP Device Locking specification

hardion:
First dig into the Device Locking system.
There is still some confusion to clear up, details to specify and reference to other RFC especially the Device Server.


piogor:
Hi @hardion, it looks like you are using a branch with the old history, so pushing README.md instead of RequestReply.md....

hardion:
Damm you are right! Is the name of the file the only difference?

piogor:
> 
> 
> Damm you are right! Is the name of the file the only difference?

Indeed, there is a difference in the content of the file, too. Missing  for example "topics points":
```
### Client duty
#### Reconnection
TODO Describe here how to check the connection and in which condition should the client reconnect

### Version compatibility
TODO describe here in which condition
TODO decide which version the specification should focus on i.e version 5

### Protocol
#### Message
SHOULD include:
* client process idendification i.e pid
* source

#### Exception

#### Timeout
#### Synchronous request
#### Asynchronous request
TODO understand the PUSH_BACK model. CORBA or more generic?

#### Cache

#### Serialisation
```

piogor:
@hardion, I have replaced this PR with #114 :-).

PR #113, opened 2020-10-22
rfc-10. #112, timeout section

piogor:
It is short but I do not know if we should go more into detaila, for example specifying that Device Server and Client SHALL use OS-level timeouts when accessing network related services/resources?

### DIFF
--- a/10/RequestReply.md
+++ b/10/RequestReply.md
@@ -79,7 +79,9 @@ SHOULD include:
 #### Timeout
 
 * On the client side, the Request-Reply protocol SHALL implement timeout mechanism for every type of request. Timeout mechanism is returning DevFailed exception if a request cannot be served in specified time called Timeout.
-* For each request the client MAY setup Timeout value or opt-out from using timeout mechanism which is equal to set Timeout to infinity.
+* For each request the client MAY setup Timeout value or opt-out from using timeout mechanism which means that:
+  - if the request is Synchronous, the client SHALL be blocked as long as it takes the request to be served. A timeout realted exception SHALL not be returned.
+  - if the request is Asynchronous, the Client SHALL NOT be blocked. The Client SHALL be able to access the request result after it is served by a Device Server without any time limits. A timeout realted exception SHALL NOT be returned. 
 * The default Timeout is 3 seconds.
 
 #### Synchronous request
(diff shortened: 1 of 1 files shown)


10/RequestReply.md
```diff
@@ -77,6 +77,11 @@ SHOULD include:
 #### Exception
 
 #### Timeout
+
+* On the client side, the Request-Reply protocol SHALL implement timeout mechanism for every type of request. Timeout mechanism is returning DevFailed expetion if a request cannot be served in specified time called Timeout.
```
bourtemb:
```suggestion
* On the client side, the Request-Reply protocol SHALL implement timeout mechanism for every type of request. Timeout mechanism is returning DevFailed exception if a request cannot be served in specified time called Timeout.
```

10/RequestReply.md
```diff
@@ -77,6 +77,11 @@ SHOULD include:
...
+* For each request the client MAY setup Timeout value or opt-out from using timeout mechanism which is equal to set Timeout to infinity.
```
bourtemb: (At least) in the C++ version, setting a timeout to 0 disables it.
1 thing important to know (in C++ at least) is that timeout involves only the requests going through the network.
If a Device Server is a client of a device exported by the same device server (same process), there are some optimizations in omniORB which is calling the methods directly without going through the network.
In this case, if the method takes more time than the configured timeout, no timeout exception will be raised and the client will be blocked until the end of the method execution.

bourtemb (commented):
I fixed a typo and added a comment on the current timeout behaviour with omniORB.

10/RequestReply.md
piogor: Regarding omniORB implementation for local call, is its behaviour (blocking) an expected one or should it be marked as kind of bug?

piogor: I will rephrase the sentence with infinity to avoid confusions in respects to how APIs define opt-out fro, timeout.

bourtemb: > 
> 
> Regarding omniORB implementation for local call, is its behaviour (blocking) an expected one or should it be marked as kind of bug?

It is expected but not well known I think. So I thought this is a good opportunity to mention this potentially misleading behaviour.
I'm actually not sure this behaviour is well documented on readthedocs.

PR #120, opened 2020-12-09
WIP: RFC-10 Cache

piogor:
This a basic specification of the cache in Tango Controls


10/RequestReply.md
```diff
@@ -155,6 +155,50 @@ A Device Server SHALL process any incoming requests in the order of their arriva
 
 #### Cache
 
+The Cache is a circular buffer of recent responses for requests genereted internally by Device Server's Polling mechanism.
```
bourtemb:
```suggestion
The Cache is a circular buffer of recent responses for requests generated internally by Device Server's Polling mechanism.
```

10/RequestReply.md
```diff
@@ -155,6 +155,50 @@ A Device Server SHALL process any incoming requests in the order of their arriva
...
+
+The Polling is periodical invoking a list of requests on Devices instantiated by a Device Server. A period of invokiong a request is called a Polling Period. 
+
+* The Cache SHALL be implemented by a Device Server.
+
+* The Polling SHALL support the following requests:
+  * read Attribute,
+  * execution of  Command with DevVoid argin,
```
bourtemb:
```suggestion
  * execution of  Command with DevVoid argin
```

10/RequestReply.md
```diff
@@ -155,6 +155,50 @@ A Device Server SHALL process any incoming requests in the order of their arriva
...
+
+* The Polling Period MAY be configured separately for each Attribute or Command of each Device spearately.
```
bourtemb:
```suggestion
* The Polling Period MAY be configured separately for each Attribute or Command of each Device.
```

10/RequestReply.md
```diff
@@ -155,6 +155,50 @@ A Device Server SHALL process any incoming requests in the order of their arriva
...
+
+* A Device Class MAY implement default settings of the Polling for Attributes and Commands.
+
+* A Client MAY change the Polling Period of any Attribute or Command of any Device.
+
+* A Client MAY switch off/on a Polling for any Attribute or Command of any Device.
+
+* The Polling SHALL be configurable by a Client via the following Admin Device commands:
+  * `AddObjPolling`:
+    It add a new object (command or attribute) to the list of object(s) to be polled. It is also with this command that the polling period is specified.
```
bourtemb:
```suggestion
    Add a new object (command or attribute) to the list of object(s) to be polled. It is also with this command that the polling period is specified.
```

10/RequestReply.md
```diff
@@ -155,6 +155,50 @@ A Device Server SHALL process any incoming requests in the order of their arriva
...
+  * `RemObjPolling`:
+    To remove one object (command or attribute) from the polled object(s) list
+  * `UpdObjPollingPeriod`:
+    Change one object polling period
+  * `StartPolling`:
+    Starts polling for the whole process
+  * `StopPolling`:
+    Stops polling for the whole process
+  * `PolledDevice`:
+    Allow a client to know which device are polled
```
bourtemb:
```suggestion
    Allow a client to know which devices are polled
```

10/RequestReply.md
```diff
@@ -155,6 +155,50 @@ A Device Server SHALL process any incoming requests in the order of their arriva
...
+  * `DevPollStatus`:
+    Allow a client to precisely knows the polling status for a device.
```
bourtemb:
```suggestion
    Allow a client to precisely know the polling status for a device.
```

10/RequestReply.md
```diff
@@ -155,6 +155,50 @@ A Device Server SHALL process any incoming requests in the order of their arriva
...
+
+* The Polling configuration defined by a Client SHALL be stored in database and applied upon a Device initialisation.
+
+* For each request from the Polling list, the Device Server SHALL store in a Cache at least the most recent response generated by the request.
+
+* The depth of the Cache (the circular buffer size) SHALL be configurable.
+  
+* For any Request, a Clinent MAY decide whether it want a response from the Cache or excute underlying request. The Device Server SHALL respond adequatly.
```
bourtemb:
```suggestion
* For any Request, a Client MAY decide whether it wants a response from the Cache or execute underlying request. The Device Server SHALL respond adequately.
```

bourtemb (commented):
I suggested some typo fixes.
It's probably because the work just started on this PR, I think some notions are still missing like the fact it is possible to fill up the cache directly from the code outside of the polling thread (Externally triggered polling).
The notion of Source for the client, which can choose when reading a command or attribute to read from the cache, from the cache with a fallback to the execution on the device if the cache is considered too old (poll_old_factor parameter in Database). It is evoked but I think the 3 possibilities should be described.
The notion of polling threads pool and the fact that it is possible to configure the number of polling threads and as well to assign some devices to the different polling threads.
I think it is important to say as well that it is possible to read the history of the polling buffer (read n last values).
The size of the circular buffer is defined with the poll ring depth setting in the DB.
If a client reads with SOURCE=CACHE and the last value in the cache is too old (last update was done before poll_old_factor * polling_period ms), an exception is sent with reason field = "API_NotUpdatedAnyMore".

piogor:
Thanks @bourtemb, yes it is just basics. However, some points I would miss without your note (the size of the buffer ;-) ) and I was about to skip the polling threads as more implementation specific. We can discuss it shortly during the kernel meeting.

